PR #195, opened 2014-11-05
Removes button to allow entering the configure allowed/blocked screen.

tombettany:


alex5imon:
what's the status of this @tombettany ?

tombettany:
I think that this was added because the screen was really confusing - but since then we updated the allowed/blocked screen to make it a little easier on the eyes. This can probably be closed without merging.

PR #204, opened 2015-01-07
Make the monitor edid parsing more resilient

tombettany:
KanoComputing/peldins#1531
The monitor edid parsing would fail if it is detected that the device is
CEA but the screen size detection fails. Adds a more resilient method to
compare the screen size. Also adds an extra check to detect if it is a
monitor or a television - if the model name includes 'TV' then it is a
good indication that the device is a television.


alex5imon:
Please merge :+1: and add to changelog

PR #209, opened 2015-01-21
Switch to use resolvconf for setting DNS

tombettany:
KanoComputing/peldins#1543
Currently, if the parental controls are enabled and the interface
obtains a new network configuration via DHCP, the parental DHCP servers
will be replaced by the DNS servers from DHCP. Resolvconf allows finer
control over which DNS servers are used, providing the ability to
prevent DNS servers configured by DHCP from being used.

This patch fixes the problem by changing the DNS settings through
resolvconf and disabling DHCP DNS servers when parental DNS is being
used.


alex5imon:
Is this ready to merge? @tombettany

tombettany:
@alex5imon Yes, with https://github.com/KanoComputing/kano-toolset/pull/77

alex5imon:
Please merge :+1:

PR #218, opened 2015-03-02
Support for pi2 clock values

Ealdwulf:
See commit messages for more explanation. Fixes #1642
Note that there are many cases that this has to support.


Ealdwulf:
@alex5imon It looks like you wrote overclock.py, do you want to review this?

alex5imon:
We might want to rename set-kano-hdmi file, since we are not just doing hdmi related stuff in there

kano_settings/boot_config.py
```diff
@@ -5,98 +5,135 @@
 # Copyright (C) 2014 Kano Computing Ltd.
```
alex5imon: addd `, 2015`

kano_settings/system/overclock_chip_support.py
```diff
@@ -0,0 +1,75 @@
+# set-kano-hdmi
```
alex5imon: change this name

alex5imon:
@Ealdwulf there are only a few minor comments from my side. Feel free to merge when you apply the changes

Ealdwulf:
Thanks! Will do.

alex5imon:
Please merge when ready :+1:

Ealdwulf:
NB I also renamed /etc/init.d/kano-hdmi to /etc/init.d/kano-settings. 
It turns out dpkg won't remove the old file kano-hdmi, but I think it is safe because the link to it in /etc/rc2.d is removed.

PR #221, opened 2015-03-03
Forcibly disabling the SYSRQ kernel keyboard hotkey

skarbat:


pazdera:
I'm not sure whether this will work on Rpi1. I think that it's compiled out of our kernel completely.

convolu:
I tested it on the Rpi2, because I think I am dealing with a related issue and it seems to be fixing the Q key issue (not mine though :frowning:  )

skarbat:
You're right @pazdera , I think it would be harmless anyway, because the /proc entry would be missing, but added a conditional to do it only if available. Looks better?

pazdera:
It would complain that the file doesn't exist. I'm happy with it under the conditional :-).

Feel free to merge :shipit:

PR #222, opened 2015-03-04
Adding an init script to play a sound during the console bootup messages

skarbat:
TODO: We need to place the sound file at `media/sounds/system_bootup.wav`

cc @alex5imon @pazdera


alex5imon:
@skarbat looks good to me! Let's wait to have the sound before merging

PR #226, opened 2015-03-09
Fixing a warning message when plugging the wifi dongle

skarbat:
* Plugging the wireless dongle will only reboot the unit
   on RaspberryPI models A and B. This change removes
   this warning message for all other models.

cc  @pazdera @FMog


alex5imon:
Please merge :+1:

PR #228, opened 2015-03-10
To allow kano-world-launcher to have a splash screen, we need to call

Ealdwulf:
kano-stop-splash from settings.
Although it would be safe for kano-world to launch without this change,
the splash would temporarily cover the UI so we bump the version
of kano-settings to allow kano-profile to depend on it.
We also need to allow the splash identifier variables to pass to
kano-settings when it is run using sudo, for which purpose the sudoers
file is modified.


Ealdwulf:
Merging as required for kano-profile now...

PR #233, opened 2015-03-13
Adding the functionality to reset Desktop settings

skarbat:
* At this point only the lxpanel settings are to be restored
 * Refreshes the panel so that no logoff is necessary


alex5imon:
Please merge :+1:

PR #235, opened 2015-03-16
Bootup sound script is displaying warning messages

skarbat:
* Missing include file for LSB init scripts (log_msg and friends)
 * Fixed order of redirection parameter for aplay

This minor issues caused garbage noise to be displayed to the boot console.

cc @alex5imon @Ealdwulf @pazdera


alex5imon:
Please merge :+1:

PR #239, opened 2015-03-26
Prevent 640 480

Ealdwulf:
This PR should fix #1723 . If we detect 640x480 mode, we use the same path as the 'safe mode' keys to reboot in 1024x768.


Ealdwulf:
If no-one has any problems with this, I'll merge it today...

PR #240, opened 2015-03-27
Moving settings to use the new kano-init API

pazdera:
Changing to the new kano-init interface.

I made all the imports conditional and while that's not ideal, there shouldn't be any problems with compatibility.

cc @alex5imon


pazdera:
The new kano-init has been merged, this is required by it, merging as well.

PR #241, opened 2015-03-30
bin: Emit a tracking event

pazdera:
This happens when the window is closed, sending off the current

  * audio configuration,
  * display setup
  * and parental lock level.

Depends on the new tracker in profile.

Related to: https://github.com/KanoComputing/peldins/issues/1661

cc @carolineclark @alex5imon 

Tested on the pi and it seems to be working.


alex5imon:
Please merge :+1:

PR #242, opened 2015-03-30
Chromium safe-search

skarbat:
* Disabling google.com when setting parental control to "high"
 * Until we have a higher version of Chromium that supports this sestting in policy files

### DIFF
--- a/kano_settings/system/advanced.py
+++ b/kano_settings/system/advanced.py
@@ -15,6 +15,8 @@ import subprocess
 import signal
 import urllib2
 from bs4 import BeautifulSoup
+import pycountry
+import gzip
 
 from kano_settings.common import settings_dir
 from kano.utils import read_file_contents, write_file_contents, \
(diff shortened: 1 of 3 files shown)


kano_settings/system/advanced.py
```diff
@@ -368,5 +368,14 @@ def set_parental_level(level_setting):
         set_ultimate_parental(False)
 
     blacklist, whitelist = read_listed_sites()
+
+    # SafeSearch: disabling google.com, until we have higher Chromium version
+    # that implements policy files.
+    if 'chromium' in enabled:
+        if blacklist:
```
alex5imon: I don't think we should do this... 

There's a function to write to the blacklist:
https://github.com/KanoComputing/kano-settings/blob/safe-search/kano_settings/system/advanced.py#L339

also blacklist should exist, if it doesn't then we have a problem we shouldn't just create the list with only 'google.com'

Also we need add more:
www.google.com
google.com
www.google.uk
google.uk
www.google.es
google.es
www.google.de
google.de

and so on...

alex5imon: check this line for the 'www' thing: https://github.com/KanoComputing/kano-settings/blob/master/kano_settings/system/advanced.py#L142-L149

skarbat:
Changed the way we disable google searches. I tested it and seems to work pretty well. @alex5imon

alex5imon:
@tombettany could you review this please ??

alex5imon:
@skarbat when testing this we will need to make sure the feedback widget still works (sending the feedback to the google doc)

alex5imon:
s***t I didn't want to close this sorry

kano_settings/system/advanced.py
```diff
@@ -15,6 +15,8 @@
 import signal
 import urllib2
 from bs4 import BeautifulSoup
+import pycountry
+import gzip
```
tombettany: Is the `gzip` module installed by default?

tombettany:
Nice solution @skarbat - especially the transferring to using the `gzip` module. Apart from my above comments, I'm happy with this.

skarbat:
That's correct @tombettany , installed on our image for Python 2.6, 2.7 and 3.2.

tombettany:
Then @skarbat, please merge :+1:  :shipit:

PR #247, opened 2015-04-20
Use alsa-utils to store HDMI vs speaker config

Ealdwulf:
Alsa-utils stores the volume and audio-route config.
We were also storing the audio route config in /etc/rc.audio and not telling alsa-utils about it.
The fix to https://github.com/KanoComputing/peldins/issues/1684 reloads the volume using alsa-utils after lxpanel is launched, this has the effect of using alsa-utils version of the audio-route config (confusingly, after the boot sound)

This PR removes our /etc/rc.audio in favour of using alsa-utils to store all of the audio config (ie, volume and HDMI vs speaker).


alex5imon:
@skarbat @pazdera can you have a look at this please? No rc.audio needed

skarbat:
It's good that we leverage this to `alsa`. Looks good to me @alex5imon @Ealdwulf

alex5imon:
Please merge :+1:

Ealdwulf:
Thanks! Merging...

PR #249, opened 2015-04-22
Remove call to /etc/rc.audio which no longer exists, and update comments...

Ealdwulf:
There was a call to /etc/rc.audio which I missed removing in #247; it doesn't break anything but produces an error message at boot. This PR removes it.


skarbat:
looks great to me @Ealdwulf, thanks!

Ealdwulf:
Thanks; merging...

PR #250, opened 2015-04-24
kano-settings should depend on kano-connect

pazdera:
cc @carolineclark


pazdera:
@Ealdwulf Yes, I forgot about this one. Thanks for bringing it up!

PR #251, opened 2015-04-24
Audio: Identify the ALSA control device by name

tombettany:
Originally, 'numid' was being used to set the PCM control value but this
can vary between systems so use the name instead. Also default to
reporting analogue if the value retrieved when checking is not
recognised.

cc @Ealdwulf


tombettany:
Fixed the thing mentioned by @Ealdwulf. Merging now.

PR #255, opened 2015-04-28
Telling kano-wifi-gui to save the connection for next boot

skarbat:
* On successfull connection the cache file is saved,
   if the connection does not succeed the cache is removed


alex5imon:
Please merge :+1:

PR #259, opened 2015-05-08
Wallpaper: Make wallpaper setting more tolerant

tombettany:
Currently the function for changing the wallpaper takes the path and the
name of the wallpaper. It then constructs the full filename by appending
the path with the name, but it does this without any checking, requiring
either the supplied path to end in a `/` or the supplied name to begin
with a `/`. Instead use `os.path.join` to make this process safer.

cc @carolineclark @convolu


convolu:
Looks good to me :shipit:

PR #260, opened 2015-05-11
Enabling ondemand mode to RPI2 4 CPU cores

skarbat:
* Even thought KanoOS kernel has it by default,
   we are forcing the "ondemand" operation mode to the rest
   of the CPU cores


pazdera:
Looks good to me :+1:

PR #262, opened 2015-05-11
Set default value for Country dropdown

convolu:
This PR sets the default country to the first one in the available list. This way after changing the continent instead of getting empty dropdown menus for country and consequently for the "advanced options"


convolu:
@alex5imon @tombettany Could you please review?

alex5imon:
Please merge :+1:

PR #263, opened 2015-05-11
Add disconnect wifi option

carolineclark:
The command ```sudo kano-wifi-gui --disconnect``` will now disconnect the user from the internet if they are connected to WiFi.

Added a disconnect option to the internet widget.
kano-screenshot-mon-may-11-17-58-30

When clicking disconnect, this launches a dialog once the process is finished:
kano-screenshot-mon-may-11-18-00-02

Added a beautiful Disconnect button to the GUI when the user is connected via WiFi.  Will change when @TommySaehl has a design, the placement does look a bit weird.
kano-screenshot-mon-may-11-18-12-31

This is nearly complete, but needs some more refining with ethernet connections - for example, the Disconnect option in the internet widget should only be shown when connected to WiFi.

Needs this pull request to work: https://github.com/KanoComputing/kano-toolset/pull/131

### DIFF
--- a/bin/kano-wifi-gui
+++ b/bin/kano-wifi-gui
@@ -9,8 +9,24 @@
 # An alternative to kano-wifi console tool.
 #
 
+
+"""
+kano-wifi-gui shows the different screens for connecting and disconnecting to
+the internet.
+
+Usage:
+  kano-wifi-gui [-d]
+  kano-wifi-gui <screen>
+Options:
+   -h, --help       Show this message.
+   -d, --disconnect
+"""
+
+
 import sys
 import os
+import docopt
+
 from gi.repository import Gtk, GObject
 
 if __name__ == '__main__' and __package__ is None:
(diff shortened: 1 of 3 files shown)


alex5imon:
@carolineclark great work! Two things:
1) Ask @mkeegan for the copy of that dialogue
2) You are right, disconnect shouldn't be an option when using Ethernet

mkeegan:
Copy:

```Disconnect complete - you're now offline.```

carolineclark:
I think this is ready to review.  @skarbat if you could review this PR, especially the C code, that would be amazing.
One thought - if connected to WiFi, we could also add a Configure option in the wifi widget, which would launch the wifi-gui.

skarbat:
the rest looks great to me @carolineclark

alex5imon:
Please merge :+1:

TommySaehl:
image
Thoughts?

alex5imon:
Seems a bit too prominent for me, seems like there is an error (more visible than the green tick)

TommySaehl:
Selecting an already established connection in the networks list should change the button from *Connect* to a red *Disconnect*.

alex5imon:
I like that

PR #264, opened 2015-05-12
debian: Removing unused dependency

pazdera:
This dependency isn't used by the package any more.

cc @Ealdwulf @skarbat @alex5imon


alex5imon:
Please merge :+1:

PR #265, opened 2015-05-18
Additional search sites added to the parental "high" mode setting

skarbat:


kano_settings/system/advanced.py
```diff
@@ -136,8 +138,37 @@ def add_safesearch_blacklist(hosts):
         ]
 
     for subdomain in second_domains:
-        hosts.append(url_pattern.format(subdomain))
-        hosts.append(www_pattern.format(subdomain))
+        hosts.append(url_pattern.format('google.' + subdomain))
+        hosts.append(www_pattern.format('google.' + subdomain))
+
+    # List of additional search sites to block
+    additional_search_sites=[
```
alex5imon: do you need to add the subdomain to these sites as well? @skarbat

kano_settings/system/advanced.py
```diff
@@ -136,8 +138,60 @@ def add_safesearch_blacklist(hosts):
...
+        ]
+
+    # Add the additional search sites
+    for additional_site in additional_search_sites:
+        hosts.append(url_pattern.format(additional_site))
+        hosts.append(www_pattern.format(additional_site))
+
+    # Add subdomains to the sites that need it
+    for country in pycountry.countries:
+        hosts.append (url_pattern.format('{}.ask.com'.format(country.alpha2.lower())))
```
alex5imon: Sorry to be annoying but can you create a function out of this code? 
```
def myfunc(mywebsite):
  hosts.append (url_pattern.format('mywebsite'.format(country.alpha2.lower())))
  hosts.append (www_pattern.format('mywebsite'.format(country.alpha2.lower())))
```

And add all the search engines into a list, that way it's much easier if in the future we need to add new ones

kano_settings/system/advanced.py
```diff
@@ -112,16 +112,18 @@ def create_empty_hosts():
...
-    www_pattern='127.0.0.1\tgoogle.{}\n'
 
+    url_pattern='127.0.0.1\t{}\n'
+    www_pattern='127.0.0.1\twww.{}\n'
+
+    # Block Google sites associated to each country (ISO 3166-1 alpha-2)
     for country in pycountry.countries:
-        hosts.append(url_pattern.format(country.alpha2.lower()))
-        hosts.append(www_pattern.format(country.alpha2.lower()))
+        hosts.append(url_pattern.format('google.{}'.format(country.alpha2.lower())))
```
alex5imon: this can be added to the other `for countr` loop, so `google` is treated as the other search engines rather than as an exception

skarbat:
Added additional cases and this looks to be working fine on the PI now @alex5imon.

PR #266, opened 2015-05-29
No Internet Screen: Fix 'later' button

tombettany:
The event connected to the 'later' button on the no internet screen was
malformed which would result in the event handler not being triggered.

cc @carolineclark @pazdera


tombettany:
@carolineclark Can I merge this?

tombettany:
@pazdera Can you take a look at this?

pazdera:
@tombettany Sorry about the delay! Looks good :shipit:

PR #267, opened 2015-06-09
i18n: Add functionality (and tests) to set locale

tombettany:
To enable use of locales, add the functions which provide the ability to
set them using the settings toolset. This adds a functions to manage the
locales (installation, removal, setting the locale, checking that the
locale is valid, etc) and also provides unit tests to check each of
these functions.

@pazdera @convolu @skarbat

### DIFF
--- a/kano_settings/system/locale.py
+++ b/kano_settings/system/locale.py
@@ -1,0 +1,149 @@
+#
+# locale.py
+#
+# Copyright (C) 2015 Kano Computing Ltd.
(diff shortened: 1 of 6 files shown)


skarbat:
Looks very clean to me @tombettany , even if not understanding the complete intrinsics. Out of interest, here is where we can add more locales on kanoos, should we need to:

 * https://github.com/KanoComputing/peldins/blob/master/fs-base/etc/locale.gen

tombettany:
@skarbat Thanks for reviewing. I'm going to merge this now as they are all new files anyway so nothing should be affected by this.

PR #268, opened 2015-06-10
Fix mislabelled call to 'logger.warning'

tombettany:
The correct call is `logger.warn`.

cc @pazdera


pazdera:
Looks good! I'll merge it right away.

PR #270, opened 2015-06-16
Home refactor

tombettany:
- Moves the management of screen icons into a separate file
 - Reimagines the home screen code structure
 - Splits the main window class from the executable
 - Allows launching screens from the CLI using either the screen name or the (to be deprecated) screen number.
 - Installs the i18n libraries

I'm not quite satisfied with the `ScreenCollection` class - we need:
 - Elements to be ordered
 - Elements accessible by name
 - Elements accessibly by index (although this will be deprecated at some point)

cc @convolu @pazdera @carolineclark Thoughts? Comments?

### DIFF
--- a/kano_settings/home_screen.py
+++ b/kano_settings/home_screen.py
@@ -2,33 +2,17 @@
 
 # home_screen.py
 #
-# Copyright (C) 2014 Kano Computing Ltd.
-# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU General Public License v2
+# Copyright (C) 2014-2015 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
 #
 
 from gi.repository import Gtk
-
-from kano_settings.set_keyboard import choose_keyboard_screen
-from kano_settings.set_mouse import SetMouse
-from kano_settings.set_notifications import SetNotifications
-from kano_settings.set_font import SetFont
-from kano_settings.set_audio import SetAudio
-from kano_settings.set_display import SetDisplay
-from kano_settings.set_wifi import SetWifi, SetProxy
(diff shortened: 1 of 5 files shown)


kano_settings/home_screen.py
```diff
@@ -36,165 +20,102 @@ class UnknownScreenError(Exception):
...
 
 class HomeScreen(Gtk.Box):
+    _DISPLAYED_SCREENS = SCREENS.get_screens_on_home()
 
-    names = ["Keyboard", "Mouse", "Audio", "Display", "Wifi", "Overclocking",
-             "Account", "Appearance", "Font", "Advanced", "About",
-             "Notifications"]
-    custom_info = ["Keyboard-country-human", "Mouse", "Audio", None, None,
-                   "Overclocking", None, "Wallpaper", "Font"]
+    def __init__(self, win, screen_number=None, screen_name=None):
```
pazdera: I'm curious why did you prefer keeping both screen number and name instead of having screen and then check for the type of the value?

pazdera:
Looks really good, @tombettany!

I left a question inline and apart of that, it's a :shipit: for me.

kano_settings/home_screen.py
tombettany: No particular reason, I suppose that I thought that it might be easier to deprecate the `screen_number` if it were a separate parameter.

If you think that it would be better otherwise then I'm happy to change it.

convolu:
@tombettany :+1: looks good to me

kano_settings/home_screen.py
pazdera: Cool. It's all right if we're removing it anyway :-).

:shipit: 

Certainly an improvement to the codebase!

tombettany:
Cool, merging now then.

PR #272, opened 2015-06-18
Allow kano-connect to proceed even if Ethernet cable

skarbat:
* When the ethernet cable is plugged in, kano-connect will still
   try to associate wirelessly if a control file is found (force multihomed mode)


tombettany:
@skarbat This looks like a good solution.

PR #273, opened 2015-06-18
Locale screen

tombettany:
Adds the Locale screen (not visible from the home screen but available with `kano-settings locale`.

Also adds a few useful locale functions

cc @convolu @pazdera @carolineclark

### DIFF
--- a/tests/i18n/test_locale.py
+++ b/tests/i18n/test_locale.py
@@ -237,3 +237,18 @@ class SetLocale(unittest.TestCase):
                     self.assertFalse(True)
 
             self.assertTrue(True)
+
+
+class GetLocale(unittest.TestCase):
+
+    def test_get_locale(self):
+        locale_param = r'^LANG='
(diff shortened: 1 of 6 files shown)


tests/i18n/test_locale.py
```diff
@@ -221,3 +237,18 @@ def test_set_locale(self):
                     self.assertFalse(True)
```
convolu: Maybe this would be more appropriate here https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertNotIn

convolu:
Looks good to me

 :+1: :+1: for unit tests. Even though not touched by this commit, there are a few places in the test where a better assert is available. Using it is useful because it allows for better feedback when the test fails.

See https://docs.python.org/2/library/unittest.html#assert-methods for reference

tombettany:
@convolu Although I agree that the asserts could be improved, I'll handle them later so I'll merge this now.

convolu:
@tombettany :+1: :shipit:

PR #274, opened 2015-06-19
Set hostname

Ealdwulf:
This PR adds a function which will set the hostname:
* On the next reboot 
* Only if it is currently 'kano'
To the name of the first user.

This is for https://github.com/KanoComputing/peldins/issues/1750
 and https://github.com/KanoComputing/peldins/issues/1893

I have tested this one by installing it from scratch. 
There will be another PR to use it from kano-init.


tombettany:
@Ealdwulf I think that (because they aren't GUI tools):
 - `kano_settings/set_hostname.py` -> `kano_settings/system/set_hostname.py`

and

 - `kano_settings/get_username.py` -> `kano_settings/system/get_username.py`

kano_settings/set_hostname.py
```diff
@@ -0,0 +1,62 @@
...
+    curr_hosts = kano.utils.read_file_contents_as_lines('/etc/hosts')
+    if '127.0.0.1\tkano' not in curr_hosts:
+        logger.warn('/etc/hosts already modified, not changing')
+
+    try:
+        lines_changed = kano.utils.sed(
+                                      '127.0.0.1\s+(kano)',
+                                      '127.0.0.1\t{}'.format(new_hostname),
+                                      '/etc/hosts',
+                                      True)
```
tombettany: @Ealdwulf I think that this might have problems because when we [set the hosts blacklist file](https://github.com/KanoComputing/kano-settings/blob/master/kano_settings/system/advanced.py#L199-L256) for parental controls, we manipulate `/etc/hosts` and, I think, completely rewrite it.

Ealdwulf: Very good point. I'll check that out.

Ealdwulf:
Okay, I've moved the modules to system.
It looks like the hosts blackllist file issue is not so bad because the existing file is at the top of
the new generated one. However we will get the old name back if parental controls are reverted. 
I will make some more changes to address this.

Ealdwulf:
@tombettany Okay, I have updated this by causing it to edit both the current hosts file and the backup hosts file.

Ealdwulf:
Since no-one objected, I'm going to merge this so it can be tested next week.

PR #276, opened 2015-06-22
Fixing get_username when Kit boots for the first time

skarbat:
When the kano kit boots for the first time, kano-init complains with `unable to resolve host kano`.
On the build logs there's an execption `KeyError: 'getgrnam(): name not found: kanousers'`.
This changeset controls this exception to return None, so the code proceeds correctly.

@Ealdwulf @alex5imon


Ealdwulf:
@skarbat Thanks for fixing this. looks good to merge from my point of view.

PR #279, opened 2015-06-25
Added monitoring to the plugin

alex5imon:
KanoComputing/peldins#1860


alex5imon:
cc @skarbat @pazdera 

The file we are monitoring takes a bit to get updated, so the change on the widget is not instant but works.

skarbat:
looks good to me @alex5imon ! Yes, detecting the "up" event takes a little time until the dhcp lease is setup, the "down" event should be faster.

PR #280, opened 2015-06-25
starting the default browser instead of chromium

skarbat:
Changes to open the default system browser, which will now be Midorid. Goes along with this changeset: https://github.com/KanoComputing/kano-toolset/pull/142

### DIFF
--- a/kano_settings/system/proxy.py
+++ b/kano_settings/system/proxy.py
@@ -5,6 +5,8 @@
 # Copyright (C) 2014 Kano Computing Ltd.
 # License: http://www.gnu.org/licenses/gpl-2.0.txt GNU General Public License v2
 #
+# This module is responsible for synchronizing proxy settings to Chromium, Midori, APT, Curl and Wget tools
+#
 
 import os
 import requests
@@ -12,12 +14,14 @@ from kano.utils import run_cmd, get_all_home_folders, delete_file, \
     write_file_contents, read_file_contents_as_lines
 
 chromium_cfg = '/etc/chromium/default'
+midori_cfg = '/etc/xdg/midori/config'
 apt_cfg = '/etc/apt/apt.conf.d/80proxy'
 
 
 def set_all_proxies(enable, host=None, port=None, username=None, password=None):
     set_apt_proxy(enable, host, port, username, password)
     set_chromium(enable, host, port)
+    set_midori(enable, host, port)
     set_curl(enable, host, port, username, password)
     set_wget(enable, host, port, username, password)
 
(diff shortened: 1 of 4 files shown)


skarbat:
cc @alex5imon @rluz

alex5imon:
Please merge :+1:

skarbat:
I have just included proxy settings to the Midori browser @alex5imon @tombettany

kano_settings/system/proxy.py
```diff
@@ -50,6 +54,63 @@ def set_chromium(enable, host=None, port=None):
...
+            config_lines.append(newitem)
+
+        return newitem
+
+    def conf_remove(item, config_lines):
+        '''
+        Remove the configuration item from the configuration file
+        '''
+        removed=False
+        for idx, line in enumerate(config_lines):
```
tombettany: You don't actually use `idx` so this line could become:

```python
for line in config_lines:
    ...
```

kano_settings/system/proxy.py
```diff
@@ -50,6 +54,63 @@ def set_chromium(enable, host=None, port=None):
...
+        newitem = '{}={}\n'.format(item, value)
+        for idx, line in enumerate(config_lines):
+            if line.startswith(item):
+                config_lines[idx]=newitem
+                found=True
+
+        if not found:
+            config_lines.append(newitem)
+
+        return newitem
```
tombettany: This function could perhaps be made easier (if you don't care about the order of config file) by just doing:

```python
def conf_add_or_replace(item, value, config_lines):
    '''
    A function to simplify editing the configuration file values
    add the item if not found, replace its value otherwise
    '''
    conf_remove(item, config_lines)

    newitem = '{}={}\n'.format(item, value)
    config_lines.append(newitem)

    return newitem
```

But if you do care about config file order then this isn't viable.

kano_settings/system/proxy.py
```diff
@@ -50,6 +54,63 @@ def set_chromium(enable, host=None, port=None):
     return
 
 
+def set_midori(enable, host=None, port=None, user_agent='Mozilla/5.0 (X11; Linux) AppleWebKit/535.22+ Midori/0.4'):
+
+    # Read Midori configuration file in
+    with open(midori_cfg, 'r') as f:
+        config_lines=f.readlines()
```
tombettany: Do we need to check that the file exists before just assuming that it will be there?

kano_settings/system/proxy.py
```diff
@@ -50,6 +54,63 @@ def set_chromium(enable, host=None, port=None):
...
+        conf_remove('http-proxy-port', config_lines)
+        conf_remove('maximum-cache-size', config_lines)
+        conf_remove('identify-as', config_lines)
+        conf_remove('user-agent', config_lines)
+
+    # Write the config file back to disk
+    with open(midori_cfg, 'w') as f:
+        f.writelines(config_lines)
+
+    return
```
tombettany: If we are not returning anything then this line is a bit pointless.

tombettany:
@skarbat Other than the comment about checking for config file existence above (the other comments are just some thoughts) I'm happy with this.

skarbat:
Thanks @tombettany ! I am merging now.

PR #284, opened 2015-06-29
Add cookie control

alex5imon:


alex5imon:
@tombettany could you have a look at this PR please?

kano_settings/system/advanced.py
```diff
@@ -490,6 +489,31 @@ def set_dns_parental(enabled):
...
 
 
+def set_youtube_cookies(enabled):
+    # The cookie enables/disables safety mode in YouTube (Midori)
+    # The .db files are located in /usr/share/kano-video
+    # It needs to be copied to /home/USERNAME/.config/midori (browser) and
+    # /home/USERNAME/.config/midori/youtube (kano-video-browser)
+
+    midori_cookie_path = '/home/{}/{}'.format(username, midori_cookie)
+    youtube_cookie_path = '/home/{}/{}'.format(username, youtube_cookie)
```
tombettany: I think that the parental mode is set for the whole system, not per user (as the [lock file] is `/etc/kano-parental-lock` and the hosts file is system-wide). This means that we'll need to copy the Midori and Youtube cookies to each home dir (unless there is a system-wide config for Midori in `/etc`?)

kano_settings/system/advanced.py
```diff
@@ -530,6 +554,7 @@ def set_parental_level(level_setting):
     else:
         set_chromium_parental('chromium' in enabled)
         set_dns_parental('dns' in enabled)
+        set_youtube_cookies('dns' in enabled)
```
tombettany: We should actually create a new feature string here (something like `'youtube-cookies'`) and add it to the [preferred setting level above](https://github.com/KanoComputing/kano-settings/blob/0359d0acde833ea8f5a52ab44fe9aed477ac3bf5/kano_settings/system/advanced.py#L533-L542), i.e. something like:

```python
     feature_levels = [
         # Low
         ['blacklist', 'youtube-cookies'],
         # Medium
         ['dns'],
         # High
         ['chromium'],
         # Ultimate
         ['ultimate']
     ]
```

kano_settings/system/advanced.py
```diff
@@ -490,6 +489,51 @@ def set_dns_parental(enabled):
...
+    # Browser: Cookie needs to be copied to /home/USERNAME/.config/midori
+    if os.path.exists(browser_safe_cookie) and \
+       os.path.exists(browser_nosafe_cookie) and \
+       os.path.exists(midori_cookie):
+        if enabled:
+            logger.debug('Enabling YouTube Safety mode for browser')
+            # Change the cookie for every user
+            for d in home_dirs:
+                midori_cookie_path = '/home/{}/{}'.format(d, midori_cookie)
+                shutil.copy(browser_safe_cookie, midori_cookie_path)
```
tombettany: Are we going to have problems with permissions if we just do a copy?

convolu:
Apart from the permissions issue raised by @tombettany and the fact that it would be cleaner to use `os.path` module for formatting paths (not a big deal though) it looks ok to me.

Ealdwulf:
It looks okay to me, providing we have decided that it's okay to overwrite all the cookies (I'm assuming they are all stored in the same db).

alex5imon:
@tombettany @convolu @Ealdwulf @pazdera permissions fixed here https://github.com/KanoComputing/kano-settings/commit/eb827f97d23dd695c4c2ac7f1e5b0c1f219b445d

Please review

PR #287, opened 2015-06-30
Make kano-settings-onboot run before lightdm.

Ealdwulf:
@alex5imon This PR foces kano-settings-onboot to run before lightdm 
It also adds some standard init boiler-plate which adds a message showing it running  in the same way as all the other scripts in init.d, and prevents it being called with 'stop'). 

However, it turns out that kano-settings-onboot takes about 15 seconds to run. Therefore maybe we don't want to run it before lightdm because doing so may prevent the desktop for 15 more seconds appearing in the case where we are not going to reboot. Thoughts @pazdera @skarbat @tombettany ?

### DIFF
--- a/etc/init.d/kano-settings
+++ b/etc/init.d/kano-settings
@@ -4,8 +4,28 @@
 # Provides:         kano-settings
 # Required-Start:
 # Required-Stop:
+# X-Start-Before:   lightdm
 # Default-Start:    2
 # Default-Stop:
 ### END INIT INFO
 
-/usr/bin/kano-settings-onboot
+. /lib/lsb/init-functions
+
+case "$1" in
+    start)
+	log_action_begin_msg "Running kano-settings-onboot"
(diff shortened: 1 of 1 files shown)


pazdera:
I'd definitely keep the boilerplate, I really like that.

With the dependency, I don't know which case is worse: the delay or unexpected reboot?

alex5imon:
We can't afford a 15 sec delay at boot. It's not that fast at the moment anyway

skarbat:
Another option is that before firing a reboot and if the Xserver is already running, a small alert could be displayed instead.

Ealdwulf:
It turns out that most of the time is spent waiting for a key press for safe boot mode, so a possibility could be to split the file so that that part can be run after lightdm but the rest is run before. There is no confusion in rebooting for safe mode after the desktop has appeared, because the users can't see it anyway. However I'm not sure we should do that before a release. @alex5imon

PR #288, opened 2015-06-30
Split kano-settings-onboot

Ealdwulf:
This PR splits kano-settings-onboot into two parts so one part can be run before lightdm while the remaining part can be run after but take more time.

### DIFF
--- a/bin/kano-safeboot-mode
+++ b/bin/kano-safeboot-mode
@@ -1,0 +1,54 @@
+#!/usr/bin/env python
+# kano-settings-onboot
+#
+# Copyright (C) 2014, 2015 Kano Computing Ltd.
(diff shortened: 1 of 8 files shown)


Ealdwulf:
Merging to allow testing tomorrow.

Ealdwulf:
@tombettany @skarbat It would probably  be a good idea to review this even though I already merged it for testing, as any bug here could be unpleasant.

tombettany:
I don't see a problem just looking through here @Ealdwulf. Hopefully all will be well.

Ealdwulf:
Thanks @tombettany

skarbat:
It looks good for me @Ealdwulf

Ealdwulf:
Thanks @skarbat

PR #296, opened 2015-08-05
Added missing wifi cache cleanup on disconnect

skarbat:
* The wireless cached credentials were not being removed on disconnect,
   so the desktop network plugin would quickly reconnect back to wireless.


alex5imon:
Please merge :+1:

PR #299, opened 2015-09-21
Use WiFi for spelling https://github.com/KanoComputing/peldins/issues…

alex5imon:
…/1975


alex5imon:
@skarbat can you review this PR please? It should just change `wifi` for `WiFi`, I had to rename the png file as well because of Settings.

skarbat:
just had a look though, looks correct to me @alex5imon

PR #300, opened 2015-09-22
Rotate

rluz:
180 Flip Screen button for the Screen kit.

Thanks for the notes @carolineclark

### DIFF
--- a/kano_settings/config_file.py
+++ b/kano_settings/config_file.py
@@ -101,4 +101,3 @@ def set_setting(variable, value):
     data[variable] = value
     write_json(settings_file, data)
     chown_path(settings_file)
-    
--- a/kano_settings/config_file.py
+++ b/kano_settings/config_file.py
@@ -41,7 +41,8 @@ defaults = {
     'Font': 'Normal',
     'Wallpaper': 'kanux-background',
     'Parental-level': 0,
-    'Locale': 'en_US'
+    'Locale': 'en_US',
+    'display_rotate': 0
 }
 
 
@@ -100,3 +101,4 @@ def set_setting(variable, value):
     data[variable] = value
     write_json(settings_file, data)
     chown_path(settings_file)
+    
--- a/kano_settings/set_display.py
+++ b/kano_settings/set_display.py
@@ -14,9 +14,9 @@ from kano_settings.components.heading import Heading
 from kano_profile.tracker import track_data
 
 import kano_settings.common as common
-from kano_settings.boot_config import set_config_comment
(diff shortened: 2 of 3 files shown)


rluz:
@alex5imon

kano_settings/set_display.py
```diff
@@ -45,7 +45,8 @@ def __init__(self, win):
         self.win.top_bar.enable_prev()
         self.win.change_prev_callback(self.win.go_to_home)
 
-        self.kano_button.set_sensitive(False)
+        #self.kano_button.set_sensitive(False)
```
alex5imon: remove comment if not needed

kano_settings/system/display.py
```diff
@@ -122,6 +122,13 @@ def set_hdmi_mode(group=None, mode=None):
 
     set_config_value("hdmi_mode", mode)
 
+# flip screen 180
+def set_flip(display_rotate=None):
```
alex5imon: what happens if the `display_rotate` variable is not found in the config file?

carolineclark: If it can't find it, it should just append the setting to the end of the /boot/config.txt file, no? https://github.com/KanoComputing/kano-settings/blob/master/kano_settings/boot_config.py#L38-L65

kano_settings/config_file.py
```diff
@@ -41,7 +41,8 @@
     'Font': 'Normal',
     'Wallpaper': 'kanux-background',
     'Parental-level': 0,
-    'Locale': 'en_US'
+    'Locale': 'en_US',
+    'display_rotate': 0
```
carolineclark: On second thoughts, this should probably be removed as we're using the boot_config setters and getters rather than the ones from config_file, so this is not kept up to date. https://github.com/KanoComputing/kano-settings/blob/master/kano_settings/config_file.py#L81-L102

kano_settings/system/display.py
rluz: @alex5imon I have tried deleting display_rotate from the boot/config.txt and run the settings again, it writes the last line of config.txt every time so no duplicates or errors

alex5imon: @rluz I think you need to use the boot/config as @carolineclark says

carolineclark: I thought he was? set_config_value is from kano_settings.boot_config https://github.com/KanoComputing/kano-settings/blob/rotate/kano_settings/system/display.py#L15

alex5imon: Sounds good! :+1:

PR #303, opened 2015-09-24
Parental: Fix setting DNS for Medium and High settings

tombettany:
When setting the parental setting to Medium or High, the ultimate
parental setting code would revert the DNS servers to Google's, not
OpenDNS's. Rectify this by running the parental DNS configuration after
the ultimate parental code so that they don't get overwritten.

@FMog Built on `scratch` for you to test.

@carolineclark @convolu Could you review.

cc @alex5imon


convolu:
Looks good to me @tombettany. I'm not very familiar with the code though. If it is not too much work I think adding the side effects of `set_ultimate_parental` to its docstring would be useful.

PR #314, opened 2015-10-19
Parental: Update blacklist 19 Oct 2015

tombettany:
Update the blacklist file for Kano OS v2.2.0 release.


tombettany:
There isn't much to review here - it is an external list anyway. I'm merging now.

PR #317, opened 2015-10-21
Enhanced website sanitation to include slashes

radujipa:
This PR includes a slightly better sanitation for supplied websites.

Apart from removing the `www.` at the beginning it also removes everything after the first `/`
e.g. `www.reddit.com/r/subreddit` becomes `reddit.com`

Related to https://github.com/KanoComputing/peldins/issues/2076

@pazdera @tombettany Could you please review this?

### DIFF
--- a/kano_settings/parental_config.py
+++ b/kano_settings/parental_config.py
@@ -154,7 +154,8 @@ class SiteList(EditableList):
 
     @staticmethod
     def _sanitise_site(site):
-        return site.replace(' ', '').lstrip('www.')
+        site = site.replace(' ', '').lstrip('www.')
+        return site.split('/', 1)[0]
 
     def _item_edited_handler(self, cellrenderertext, path, new_text):
         site = self._sanitise_site(new_text)
--- a/kano_settings/parental_config.py
+++ b/kano_settings/parental_config.py
@@ -6,6 +6,9 @@
 # License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
 #
 
+
+from urlparse import urlparse
+
 from gi.repository import Gtk
 
 from kano_settings import common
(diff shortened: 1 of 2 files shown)


convolu:
Why not use https://docs.python.org/2/library/urlparse.html? it is in the standard library

radujipa:
@convolu I added `urlparse` and made the sanitation a bit better.

Also, I've added a dialog explaining why adding a link might fail.

kano_settings/parental_config.py
```diff
@@ -153,8 +156,19 @@ def __init__(self, size_x=250, size_y=25):
         EditableList.__init__(self, size_x=size_x, size_y=size_y)
 
     @staticmethod
-    def _sanitise_site(site):
-        return site.replace(' ', '').lstrip('www.')
+    def _sanitise_site(url):
+        parse = urlparse(url)
+
+        # enforcing some standards (e.g. http://)
+        if parse.scheme == '' or parse.netloc == '':
```
tombettany: Doing

```python
urlparse('www.google.com') = ParseResult(
    scheme='',
    netloc='',
    path='www.google.com',
    params='',
    query='',
    fragment=''
)
```

This `if` would cause failure.

radujipa: This is expected. We need people to enter the full URL. See dialog copy.

tombettany: I would aaargh-ue that the protocol used isn't relevant anyway, just the hostname. We should try to manipulate whatever they throw at us into the correct form rather than require that they enter it in the correct form to begin with :anchor:

radujipa: I see your point and true, the protocol isn't important. However, as per your example, the `netloc` goes into the `path` field. That means that `www.reddit.com/r/subreddit` would be in the path. I don't believe a simple url sanitation needs to turn into a full blown url validator that would require extensive testing with a little effort on their part.

radujipa:
bump

kano_settings/parental_config.py
tombettany: I don't think that we should be making this harder for the user than it already is. @convolu @convolu what do you think?

convolu: While I do agree that it requires slightly more effort from the parent, feedback is instant and thus errors are much easier to fix

convolu:
@alex5imon Could you please advise on this?

alex5imon:
I see a point on both sides... what I would say right now is to wait and see what parents want, rather than us guessing. Let's go with what we have, and change it if we get complains/suggestions

alex5imon:
cc @convolu @radujipa @tombettany

convolu:
I think that this PR attempts to fix a bug where the first 'w' characters are removed when the provided url doesn't start with `www.`. Even if we don't adopt this solution, which I think is good enough, we need to at least provide better feedback on the format of the input.

tombettany:
Okay, for the record I believe that you could use the regex:

```python
import re
regex = re.compile(ur'^(?:.*:\/\/)?(?:www\.)?([\-0-9a-zA-Z\.]+)(?:$|\/.*$)')
urls = regex.findall(test_str)
```

but stick with what you have now

tombettany:
:shipit:

PR #319, opened 2015-10-22
Suppress the output of fsck for /boot in the same way as for /

Ealdwulf:
@skarbat Here is the change we discussed.


skarbat:
@Ealdwulf tested your additions locally and they work well. good to merge.

PR #326, opened 2015-10-28
Replaced shutdown/reboot with systemctl

radujipa:
**NOTE**: how the `try` indenting in `set_account` not made it explode is beyond me

related to https://github.com/KanoComputing/peldins/issues/2122


skarbat:
again, we need a sudoers rule for `systemctl`, unless I am missing it . Perhaps here is a good choice @radujipa ?

 * https://github.com/KanoComputing/kano-settings/blob/master/debian/kano-settings.postinst#L19

radujipa:
@skarbat same mention about this here:
 https://github.com/KanoComputing/kano-apps/pull/69#issuecomment-151865093

PR #327, opened 2015-10-29
Fixed random exceptions on particular displays

skarbat:
* Some displays may return garbage data when querying EDID info
   on the model name data field. Fixed by sanitizing the returned string.

cc @Ealdwulf @alex5imon

### DIFF
--- a/kano_settings/system/display.py
+++ b/kano_settings/system/display.py
@@ -298,7 +298,9 @@ def parse_edid(edid_txt):
 
         # model name
         elif 'monitor name is' in l:
-            edid['model'] = l.split('monitor name is')[1].strip()
+            # Some displays return garbage, on old firmwares it can also be random.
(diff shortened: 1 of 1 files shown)


Ealdwulf:
This looks good @skarbat. One thing is that it's not clear to me from the traceback in the ticket that it's definitley in *that* case the monitor name that contains the junk character, although it clearly is in the case of your screen. Do you think it's worth sanitising the other data we extract from edid?

alex5imon:
@Ealdwulf @skarbat looks good. Shall we include this on Jessie?

Ealdwulf:
@alex5imon If you think it's worth doing another build (or we are doing one anyway).

pazdera:
We're definitely doing an another build :-)

skarbat:
Thanks @Ealdwulf , this is the only field that I have seen returning garbage, but worth improving this later. merging.

PR #329, opened 2015-12-16
Option in Notifications to turn ON/OFF the LED Speaker CPU Monitor Animation

radujipa:


radujipa:
https://github.com/KanoComputing/peldins/issues/2146

kano_settings/set_notifications.py
```diff
@@ -36,10 +41,19 @@ def __init__(self, win):
         self.enable_all_radiobutton = self.get_button(0)
         self.disable_all_radiobutton = self.get_button(1)
         self.disable_world_radiobutton = self.get_button(2)
-        self.show_configuration()
+
+        if is_model_2_b() is True:
```
convolu: `is True` is redundant, I think

radujipa: Good spot, thank you!

radujipa:
I've changed the behaviour of the checkbox slightly here.

 - rpi1: there is no checkbox, feature available for rp2 and upwards (OFF by default)
 - rpi2: checkbox only shows when the LED Speaker is plugged in (ON by default)

tombettany:
:+1:

PR #331, opened 2016-01-07
Parental: Prevent hanging when starting sentry server

tombettany:
KanoComputing/peldins#2175
When attempting to connect to WiFi, with ultimate parental controls
enabled and the sentry server not running, the GUI can hang despite
having fully established a connection. This is because launching the
sentry server was done in a way that it was waiting on stdin. Resolve
this by explicitly declaring the stdin source.

Supersedes https://github.com/KanoComputing/kano-settings/pull/330

@convolu @Ealdwulf Can we try this again?

### DIFF
--- a/kano_settings/system/advanced.py
+++ b/kano_settings/system/advanced.py
@@ -430,7 +430,10 @@ def redirect_traffic_to_localhost():
 
 
 def launch_sentry_server(filename):
-    subprocess.Popen(["sentry -c {}".format(filename)], shell=True)
+    subprocess.Popen(
+        ["sentry -c {}".format(filename)], shell=True,
+        stdin=subprocess.PIPE, stdout=subprocess.PIPE
+    )
 
 
 def kill_server():


tombettany:
@convolu @Ealdwulf Bump

kano_settings/system/advanced.py
```diff
@@ -430,7 +430,10 @@ def redirect_traffic_to_localhost():
 
 
 def launch_sentry_server(filename):
-    subprocess.Popen(["sentry -c {}".format(filename)], shell=True)
+    subprocess.Popen(
```
convolu: Why not use `run_cmd` or `run_bg`?

tombettany: @convolu That is a good question and the only reason that I have is that presumably it wasn't used originally for some reason.

Ealdwulf: Don't know the specific reason here, but I'm a bit wary personally of using run_cmd as it can be rather slow.

Ealdwulf:
Looks good to me...

kano_settings/system/advanced.py
convolu: I guess one would use it for the convenience of its optional parameters, or run_cmd_log to log if need be. We probably wouldn't have this issue in the first place if we used it though.

Indeed importing the whole of kano utils adds to the lag.

convolu:
:shipit:

PR #333, opened 2016-01-07
Add a dry-run mode to kano-settings-onboot (and necessarily, boot_con…

Ealdwulf:
This PR adds a --dry-run option to kano-settings-onboot which causes it to do nothing and to log all config changes and reboots.
Rebased (only change is adding dry-run logic to new  remove_noobs_defaults())
@skarbat @tombettany


skarbat:
looks ok to me @Ealdwulf !

PR #334, opened 2016-01-07
Limit reboots

Ealdwulf:
@alex5imon @convolu 
This PR prevents an unlimited number of reboots. It contains:
* A wrapper script for systemctl reboot. This builds up sentinel files per reboot and refuses to work
if there are too many. It also does sync itself, so I have removed that from call locations.
* An init script which clears the sentinels, when it appears that the system has settled.

It appears to be unavoiable that we have a timeout-based heuristc to see if teh system has settled. 
I have made this 5 minutes. That means that if someone changes there settings quickly, it won't have an effect due to refusing a reboot. Any opinions on the correct period? @FMog, I guess you are most likely to be bitten by this! 

I have set the max number of reboots at 3 (because I know kano-init reset can cause this scenario). Anyone think a different figure should apply, please comment.

This is for https://github.com/KanoComputing/peldins/issues/2186

### DIFF
--- a/etc/init.d/kano-reboot-clear
+++ b/etc/init.d/kano-reboot-clear
@@ -25,9 +25,7 @@ case "$1" in
         # Display a welcome message to the bootup terminal
 	log_action_begin_msg "Running kano-reboot-clear"
 	(
-	    echo "kano-reboot-clear starting" >/dev/tty1
 	    sleep 333 ;	rm -rf /boot/kano_reboots	
-	    echo "kano-reboot-clear done" >/dev/tty1
 	) & 
 	log_action_end_msg $?
 	;;
--- a/etc/init.d/kano-reboot-clear
+++ b/etc/init.d/kano-reboot-clear
@@ -1,4 +1,11 @@
 #!/bin/sh
+# 
+# kano-reboot-clear
+#
+# Copyright (C) 2016 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU General Public License v2
+#
+
 
 ### BEGIN INIT INFO
 # Provides:         kano-reboot-clear
--- a/etc/init.d/kano-reboot-clear
+++ b/etc/init.d/kano-reboot-clear
@@ -1,0 +1,35 @@
+#!/bin/sh
+
+### BEGIN INIT INFO
+# Provides:         kano-reboot-clear
+# Required-Start:   $local_fs $all kano-settings kano-safeboot-mode
+# Required-Stop:
+# X-Start-Before:   
+# Default-Start:    2 3 4 5
+# Default-Stop:
+### END INIT INFO
+
+. /lib/lsb/init-functions
+
+case "$1" in
+    start)
+        # Display a welcome message to the bootup terminal
+	log_action_begin_msg "Running kano-reboot-clear"
+	echo "kano-reboot-clear starting" >/dev/tty1
+	sleep 333
+	echo "kano-reboot-clear starting" >/dev/tty1 
+	rm -rf /boot/kano_reboots	
+	log_action_end_msg $?
+	;;
+    stop)
+	;;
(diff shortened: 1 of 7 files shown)


convolu:
Nice one @Ealdwulf, I especially like the fact that all these reboot files are tagged with the reason and reside in `/boot` so that we can inspect them on bad SDs. :+1: 

Maybe adding a timestamp to the files could help to figure out the order of the reboots, to help debugging. Do you think it would be a good addition?

Do you think there is ever the chance of a permission problem given the permissions on `/boot `. I assume that this isn't a problem for the places where it is being used at the moment (given it replaces the `systemctl reboot` which also requires elevated privileges?).

Regarding the delay before removing the reboot files (`333`), it seems a bit too large, but I don't have a good argument to back this :-)

etc/init.d/kano-reboot-clear
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
+
```
alex5imon: needs the Kano copyright :)

alex5imon:
@Ealdwulf "In particular, the case where you have done kano-init-reset, which currently forces 3 reboots (reboot to reconfigure screen, reboot to restore default config, kano-init-reset, reboot to configure screen again, should work. (IE, the reset should have happened, and correct final screen resolution should be present)."

Are those 3 reboots strictly needed? Could we get away with only 2?

convolu:
Would something based on this http://unix.stackexchange.com/questions/39226/how-to-run-a-script-with-systemd-right-before-shutdown help us remove the files more reliably at the right time?

etc/init.d/kano-reboot-clear
Ealdwulf: Added. NB, I've checked that systemd doesn't care where the LSB block is in the file (so it doesn't matter that the copyright is above it) but this shouldn't be merged back to wheezy without doing the same check in sysvinit.

Ealdwulf:
Thanks @convolu. I could add a timestamp, but I think it would not be valid because the network is not up yet.  Especially in the case of wifi.

I think that there shoudl not be a permission problem, because as you say, it is run as root. We could run out of space which could be avoided by having a counter, but it seems to me that adding files is more robust under normal conditions than a counter. 

Deleting the files on shutdown is an interesting idea. However, it would require us to do `reboot --force` whenever we do a reboot during bootup - otherwise the files would be deleted then, defeating the object. safeboot and settings use `--force`, but kano-init does not at this point. We could make it use `--force`, but   I am reluctant to commit to requiring `--force` because I have a suspicion that it might be what is causing config.txt to be corrupted, causing this problem (https://github.com/KanoComputing/peldins/issues/2170)in the first place!

Ealdwulf:
@alex5imon , we could in principle get away with ony 2 reboots. I guess we coudl do this by making kano-settings-onboot check for kano_init Status.RESET_STAGE and avoid rebooting in that case.
Seems a little complicated though for something that's quite rare except under test..

alex5imon:
fair enough @Ealdwulf

convolu:
Thanks @Ealdwulf, regarding the timestamps I thought it would be interesting to know primarily the order of the reboots, as identified by their tags (`safeboot`, `changes`, `reboot_now`) rather than the absolute time of them happening. Unless network comes up in between and turns this on its head.

I think you are right regarding the permissions, I only to double check (we have been affected many times by permission errors :-).

Instead of the shutdown target, I was wondering whether there is an equivalent to a reboot target on systemd we could tie this script to so that we don't have to modify our calls. This was mostly an idea, I don't know systemd as well to be sure of its feasibility

Ealdwulf:
@convolu I don't think we can know the order of the reboots, if the network is not up, because the time will be reset each boot, won't it?

I would like to merge this for testing, anyone think I should still persue a better approach?

convolu:
I'm happy with it :shipit:

PR #335, opened 2016-01-13
Parental: Ensure parental setting gets updated

tombettany:
If the ultimate parental control has been set and disabled without
updating the config file, then the script that starts the Sentry server
reports that parental control is enabled and thus network settings don't
get properly set. Resolve this by writing the change to the config file
every time some change occurs to the parental setting.

cc @convolu @Ealdwulf


tombettany:
Anyone any problems here?

tombettany:
Merging this now.

PR #336, opened 2016-01-15
Config locking

Ealdwulf:
This PR changes the settings API to be transactional, and changes the settings tools to use it.

All changes are applied to temporary file in `/boot`, which is moved into place at the end. This will hopefullly prevent the 'empty config file' problem.

All the functions which write  to `/boot/config.txt` will no longer take effect until a call to end_config_transaction(). This includes things like `safe_mode_restore_config`as well as the more obvious ones.  Note that read functions as well as write functions start a transaction. During a transaction, a lock is held so any access by other processes will wait (for up to 5 seconds) and then time out. This wait is in case the boot scripts happen to run at the same time.

This is a bit a bit complicated so APIs used externally to  the kano-settings repo  do an entire transaction. Fortunately there are not many of these. So while some scripts use `set_config_value` and the new `end_config_transaction`, these are ones that live in the kano-settings repo (eg kano-settings-onboot). For this purpose I've moved the config setting in kano-init into a new API function kano_settings.default.set_default_config(), and made the one API used in kano-init-flow (`write_overscan values`) terminate its transaction, and similarly `is_overscan()` which is used by kano-video also terminates its transaction.

@skarbat @pazdera @convolu @tombettany

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -19,7 +19,8 @@ sys.path.insert(1, os.path.abspath(
 
 SUITE = unittest.TestSuite()
 TESTS = [
-    'tests.i18n.test_locale'
+#    'tests.i18n.test_locale',
+    'tests.boot_config.test_config'
 ]
 
 for test in TESTS:
--- a/tests/test.py
+++ b/tests/test.py
@@ -19,7 +19,7 @@ sys.path.insert(1, os.path.abspath(
 
 SUITE = unittest.TestSuite()
 TESTS = [
-#    'tests.i18n.test_locale',
+    'tests.i18n.test_locale',
     'tests.boot_config.test_config'
 ]
 
--- a/bin/kano-safeboot-mode
+++ b/bin/kano-safeboot-mode
@@ -17,7 +17,8 @@ if __name__ == '__main__' and __package__ is None:
 
 from kano.utils import run_cmd, enforce_root
 from kano.logging import logger
-from kano_settings.boot_config import enforce_pi, is_safe_boot, safe_mode_backup_config
+from kano_settings.boot_config import enforce_pi, is_safe_boot, \
+    safe_mode_backup_config, end_config_transaction
 from kano_settings.system.display import set_safeboot_mode
 
 logger.force_log_level('info')
@@ -47,6 +48,8 @@ if safe_boot_requested() and not is_safe_boot():
     safe_mode_backup_config()
 
     set_safeboot_mode()
+
+    end_config_transaction()
 
     # Trigger a reboot
     logger.sync()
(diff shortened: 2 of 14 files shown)


Ealdwulf:
@skarbat This PR also has some tests. I have only run them manually so possibly they will not run correctly under the QA server.

convolu:
Looks ok to me @Ealdwulf. I think you have captured all the uses of functions that may be reading or editing the config. We will need to be careful to ensure we end the transactions in the future if we are doing anything related.

Do you think that adding a call to the `end_config_transaction()` to the `atexit` hook if one imports `kano_settings/boot_config.py` would make it safer? If it may be introducing more complexity, we can avoid it for now.

tests/test.py
```diff
@@ -19,7 +19,8 @@
 
 SUITE = unittest.TestSuite()
 TESTS = [
-    'tests.i18n.test_locale'
+#    'tests.i18n.test_locale',
```
tombettany: Is there a reason to disable the i18n tests?

Ealdwulf: Good catch! No there isn't, I just commented them out because I only wanted to run mine. Will fix.

tombettany:
Otherwise looks good :shipit:

Ealdwulf:
I was wondering about using `atexit` @convolu . I left it out on the grounds that otherwise we won't notice that we've left out an `end_config_transaction()`, but I don't really feel strongly about that. I don't think it would add much complexity because `end_config_transaction()` should already be a no-op if there isn't one. If the consensus is that `atexit` would be a good idea, I'll put it in. Opinions @tombettany @pazdera @skarbat ?

pazdera:
Looks good :+1:. Just curious: what happens if we forget to end the transaction? Is that easily debuggable?

Ealdwulf:
@pazdera At the moment it will silently not write back any changes. It won't leave the file locked, as linux will close the lock file when the process ends. Do you think we should have an `atexit` entry to close the transaction as @convolu suggests? Or alternatively, an `atexit` entry that just complains if there is an open transaction.

tombettany:
An `atexit` sounds reasonable. It should probably close the transaction - just complaining will probably go unnoticed and not really help us.

pazdera:
@Ealdwulf @tombettany @convolu The at_exit sounds good. I'd close the transaction and crash the process with an exception to make it clear there's an error.

Ealdwulf:
Okay, I've added an atexit handler, closing the transaction and raising an exception if there was one. . (BTW, It turns out that python will even catch the exception and run any remaining atexit handlers before raising it again). Will now merge for testing.

PR #338, opened 2016-01-20
wallpapers: Adding a getter to the wallpaper API

pazdera:
We need to set the background in the init flow based on what the user chooses in the picker. This will allow us to do that.

Related to: https://github.com/KanoComputing/peldins/issues/2212

cc @alex5imon @tombettany @convolu @skarbat

### DIFF
--- a/kano_settings/system/wallpaper.py
+++ b/kano_settings/system/wallpaper.py
@@ -36,14 +36,17 @@ def _get_wallpaper_from_kdeskrc(kdeskrc_path):
     """
 
     wallpapers = {}
-    if os.path.isfile(kdeskrc_path):
-        with open(kdeskrc_path, 'r') as kdesk_conf:
-            for kdesk_conf_line in kdesk_conf:
-                for key, suffix in SIZE_SUFFIX_MAP.iteritems():
-                    token = "{}:".format(key)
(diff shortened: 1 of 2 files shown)


skarbat:
I like that one @pazdera !

kano_settings/system/wallpaper.py
```diff
@@ -2,36 +2,79 @@
...
+
+    wallpapers = {}
+    if os.path.isfile(kdeskrc_path):
+        with open(kdeskrc_path, 'r') as kdesk_conf:
+            for kdesk_conf_line in kdesk_conf:
+                for key, suffix in SIZE_SUFFIX_MAP.iteritems():
+                    token = "{}:".format(key)
+                    if token in kdesk_conf_line:
+                        path = kdesk_conf_line.split(token)[-1].strip()
+                        wallpapers[key] = path
```
tombettany: This looks a bit Zsolty but it does seem to do the job and I can't find an easy solution to the problem (other than creating a function for something like `parse_kdesk_conf_line(conf_line)`)

tombettany:
Other than comment above, looks alright

convolu:
Looks ok, :shipit:

PR #340, opened 2016-01-20
Correct corrupt config

Ealdwulf:
@alex5imon @skarbat @pazdera 
This PR is for https://github.com/KanoComputing/peldins/issues/2170
It performs simple (not exhaustive) checks on config.txt and restores a default one if anything alarming is seen. Currently, it only checks for the presence of two parameters chosen from different locations  in the file.
This PR makes a copy of the config.txt from peldins. To avoid divergence, I will later make a PR for peldins which copies this file into `/boot` during the build, after kano-setttings has been installed.

### DIFF
--- a/bin/kano-settings-onboot
+++ b/bin/kano-settings-onboot
@@ -162,6 +162,7 @@ enforce_root('Need to be root!')
 
 # Check for corrupt config file
 if check_corrupt_config():
+    logger.error("Found corrupt config file! Restoring default.")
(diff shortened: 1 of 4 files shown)


skarbat:
Very detailed work @Ealdwulf , seems ok to me.

Ealdwulf:
Thanks @skarbat . I've added a log and will now merge for testing...

PR #341, opened 2016-01-22
Enable spi device by default.

Ealdwulf:
@skarbat @pazdera @alex5imon 
This enables the spi device by default.


Ealdwulf:
Merging for testing

PR #343, opened 2016-01-27
Increase the overscan limits

pazdera:
Related to: https://github.com/KanoComputing/peldins/issues/2197

The overscan limits in settings were set to 100. This change makes it dynamic based on the system resolution (maximum cutoff is 1/3 of the screen at each edge).

cc @alex5imon @convolu @Ealdwulf @skarbat @tombettany


skarbat:
looks pretty good to me @pazdera Radek

PR #346, opened 2016-02-03
Parental: Permit setting lowest setting

tombettany:
KanoComputing/peldins#2270
When setting the parental level, the GUI setting is saved also with a
prevention mechanism for setting the value too low. This minimum was
maligned with the numbering for the parental setting which prevented the
lowest setting from being configured. Resolve this by correcting the
minimum setting.

cc @pazdera

### DIFF
--- a/kano_settings/system/advanced.py
+++ b/kano_settings/system/advanced.py
@@ -586,7 +586,7 @@ def write_blacklisted_sites(blacklist):
 
 
 def set_parental_level(level_setting):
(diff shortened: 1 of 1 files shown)


pazdera:
:shipit:

pazdera:
Can you also push to devel when you do please? I'll burn today's image to test.

tombettany:
@pazdera It should be in `devel` now.

pazdera:
Thanks @tombettany !

PR #347, opened 2016-02-03
Change remove_noobs_defaults() to only move to r/w state when it actually needs to

Ealdwulf:
This PR  fixed the traceback seen from kano-settings-onboot. This was logically safe, but would cause unnecessary  writes to `/boot/config.txt` which should be avoided (as well as noise in the logs).
@convolu @tombettany


Ealdwulf:
@pazdera I want to get this into the build as well, can you review?

convolu:
Looks good to me :+1:

Ealdwulf:
Thanks @convolu

PR #350, opened 2016-02-10
WiFi: Rely on libs module for wlan interface

tombettany:
To support variable network interfaces, don't hard-code the network
interface to use.

To be merged after release of v2.3.0.

cc @Ealdwulf @convolu @radujipa

### DIFF
--- a/etc/network/interfaces
+++ b/etc/network/interfaces
@@ -14,8 +14,16 @@ iface eth0 inet manual
  post-up /sbin/udhcpc -p /var/run/udhcpc-eth0.pid --script=/etc/udhcpc/kano.script -S -i eth0
  pre-down pkill -F /var/run/udhcpc-eth0.pid 2>/dev/null || /bin/true
 
+allow-hotplug eth1
+iface eth1 inet manual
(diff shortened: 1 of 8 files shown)


etc/network/interfaces
```diff
@@ -14,9 +14,17 @@ iface eth0 inet manual
  post-up /sbin/udhcpc -p /var/run/udhcpc-eth0.pid --script=/etc/udhcpc/kano.script -S -i eth0
  pre-down pkill -F /var/run/udhcpc-eth0.pid 2>/dev/null || /bin/true
 
+allow-hotplug eth1
+iface eth0 inet manual
+ post-up /sbin/udhcpc -p /var/run/udhcpc-eth0.pid --script=/etc/udhcpc/kano.script -S -i eth1
+ pre-down pkill -F /var/run/udhcpc-eth0.pid 2>/dev/null || /bin/true
```
convolu: Is this eth0 right here?

convolu: Oh I see, disregard the above message.

convolu:
Looks ok to me

etc/network/interfaces
skarbat: I think you were right @convolu , both eth lines should match the same interface?

tombettany:
Merging for testing.

PR #352, opened 2016-02-16
Add methods to switch to/from the experimental GL graphics driver.

Ealdwulf:
This PR adds an a way to switch to the new driver, if it is installed, for testing.
There will be another PR for kano-desktop to start xcompmgr if necessary.
It saves `/usr/share/X11/xorg.conf.d/99-fbturbo.conf`, which is actually from `xserver-xorg-video-fbturbo`, in `/var/cache`, and only restores it if there is not one there already (therefore hopefully avoiding overwriting it if it is written by upgrading that package).
@tombettany @radujipa @alex5imon


tombettany:
Looks okay to me :+1:

Ealdwulf:
Thanks @tombettany , merging...

PR #355, opened 2016-02-18
Boards: Make detection systems hardware agnostic

tombettany:
Abstract board settings to make it simple to support boards with other
settings.

Requires: https://github.com/KanoComputing/kano-toolset/pull/193

First draft and not tested yet but what do you guys think of the approach?

@convolu @radujipa @Ealdwulf

### DIFF
--- a/kano_settings/system/boards/rpi_1.py
+++ b/kano_settings/system/boards/rpi_1.py
@@ -1,0 +1,68 @@
+#
+# rpi_1.py
+#
+# Copyright (C) 2016 Kano Computing Ltd.
+# License: http://www.gnu.org/licenses/gpl-2.0.txt GNU GPL v2
+#
+# Board specific settings for Raspberry Pi 1 (A/A+/B/B+)
+# TODO: Should this support the Compute Module and RPi Zero also or should
+#       another module be created which just links back to these settings?
+#
+
+DEFAULT_CONFIG = {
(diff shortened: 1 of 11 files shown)


kano_settings/system/boards/rpi_1.py
```diff
@@ -0,0 +1,68 @@
...
+    'Wifi-connection-attempted': False,
+    'Overclocking': 'High',
+    'Mouse': 'Normal',
+    'Font': 'Normal',
+    'Wallpaper': 'kanux-background',
+    'Parental-level': 0,
+    'Locale': 'en_US',
+    'LED-Speaker-anim': False,
+    'Use_GLX': False
+}
```
Ealdwulf: Do we need all of these things in the board config? Eg, the wallpaper, font etc don't seem to be related to the board.

tombettany: Fair point - it really is just the default overclock setting which is board specific.

Ealdwulf: And LED-speaker-anim. But otherwise yes.

Ealdwulf:
Apart from the comment above this looks like a great simplification :+1:

tombettany:
@Ealdwulf fixed your comment above. I'll merge for us to test.

PR #358, opened 2016-02-23
Switch browser

Ealdwulf:
This PR sets x-www-browser to epiphany on pi1 and pi0, and chromium otherwise.
Note that currently it is set to epiphany everywhere but we hardcode chromium. There will be another PR in kano-desktop to switch it in kdesk.
@alex5imon @tombettany


tombettany:
Looks good to me :shipit:

Ealdwulf:
Thanks @tombettany, merging...

PR #360, opened 2016-03-01
Fix low level parental control

Ealdwulf:
This PR is intended to fix https://github.com/KanoComputing/peldins/issues/2294. It needs more testing.
Making a PR only to remember why this branch exists and link it to the issue.
 - sentry dns is now used for low level as well as ultimate level.
 - In low level, sentry is used to map *.(google|youtube).*DOMAIN to
   forcesafesearch.google.com
 - Cookie manipulation is no longer required
This depends on a patched version of sentry (version 0.5.1) (https://github.com/KanoComputing/sentry/pull/2)


Ealdwulf:
As discussed in scrum, merging for testing.

PR #364, opened 2016-03-24
Added CLI option to set overscan values

skarbat:
@tombettany this will allow us to set the new overscan values persistently.


tombettany:
Looks good

PR #365, opened 2016-03-30
Revert gpu mem increase.

Ealdwulf:
This PR reverts the gpu memory increase. Anyone wanting to make a case for 192 please do so, otherwise  will commit this - we have 31Mb of headroom which ought to be enough, although we shoudl keep an eye on it.
@radujipa @alex5imon


Ealdwulf:
Was putting off merging this in case the video used lots of memory, but we are not using that so I will merge it.

PR #366, opened 2016-04-01
Calling Dashboard Onboot script early in the boot process

skarbat:
* The script ensures user is always brought to the Dashboard
 * Using a safe soft dependency to the external script


Ealdwulf:
Makes sense :+1:

skarbat:
Thanks @Ealdwulf , merging.

PR #367, opened 2016-04-11
force_hdmi_hotplug=1 appears to cause problems on pi3 and

Ealdwulf:
This appears solve https://github.com/KanoComputing/peldins/issues/2315 but I want to run some more tests overnight to see if it completely solves it.


Ealdwulf:
Unfortunately the pi foundation don't believe this will solve the problem: hhttps://github.com/raspberrypi/firmware/issues/585#issuecomment-208467571

PR #368, opened 2016-04-12
CLI: Implement setting audio

tombettany:
The CLI tool already included a command to set the audio mode however
this was unimplemented and hence did nothing when called. Connect this
command with the associated action.

cc @skarbat @Ealdwulf


skarbat:
Oooh! sorry about that @tombettany . Thanks for fixing! @tombettany

PR #370, opened 2016-04-19
Fixes for tvservice

skarbat:
* Avoid calling tvservice when the audio module is imported
 * Clarified code, return a meaningul response after setting audio output

### DIFF
--- a/kano_settings/system/audio.py
+++ b/kano_settings/system/audio.py
@@ -29,16 +29,29 @@ analogue_cmd = "amixer -c 0 cset numid=3 1"
 hdmi_cmd = "amixer -c 0 cset numid=3 2"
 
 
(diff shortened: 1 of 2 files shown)


skarbat:
@Ealdwulf what do you think? Tested on the RPi with a patched `tvservice` and it doesn't get called at load time anymore.

The paths fix I provided might not have system wise implications me thinks. They are symlinked.

Ealdwulf:
Looks good :shipit:

skarbat:
Thanks @Ealdwulf , merging in

Ealdwulf:
Added to changelog for 3.1.0 Haven't seen it crash after switching back and forth a bunch of times, so marking as done.

PR #372, opened 2016-04-20
Move a slow import to where it is used.

Ealdwulf:
This PR moves an import to speed up kano-settings-cli
@alex5imon @tombettany
This is for KanoComputing/kano-overworld#215, as if kano-settings-cli can be made fast enough, we don't need to use vcgencmd.


alex5imon:
Please merge 👍

PR #373, opened 2016-04-21
Reverting back a missing function name

skarbat:
* On a previous commit, I mistakenly renamed a function
   which was used by kano-settings GUI app, and possibly
   we use it elsewhere on the system.

cc @Ealdwulf


Ealdwulf:
Yeah I notices that too @skarbat . But I think this is not right, because it was previously a boolean, not a function. I've made a fix here: https://github.com/KanoComputing/kano-settings/pull/374

skarbat:
Sorry about that @Ealdwulf ! I'll close this one so yours takes precedence.

PR #374, opened 2016-04-21
Fix broken import. Also avoid calling tvservice more than once.

Ealdwulf:
Fixes broken import and caches result of is_hdmi_audio_import()
@skarbat


skarbat:
Looks good to me @Ealdwulf . Thanks

Ealdwulf:
Thanks @skarbat , merging...

PR #375, opened 2016-04-27
Fixing kano-settings keyboard sometimes failing

skarbat:
* A thread was started to apply the new changes (read), and right
   after that a call to save them was being called (write),
   the apply would sometimes crash because the file would be empty.

 * Fixed by saving the settings first, then start the thread to apply them.

@Ealdwulf this PR would replace this one: https://github.com/KanoComputing/kano-toolset/pull/208

Tested briefly and seems to fix it. What do you think?


Ealdwulf:
Much nicer @skarbat :+1:

skarbat:
Thanks @Ealdwulf !

PR #376, opened 2016-04-27
Limit display resolutions

skarbat:
* Filter out those below 1024x768 so the user cannot select them
   from the kano-settings Display section.
 * Should fix problems when selecting a lower resolution which is then
   automatically increased on next reboot.


Ealdwulf:
Looks good :+1:

PR #379, opened 2016-05-04
Dependencies: Depend on new libkano-networking package

tombettany:
The `libkano_networking` library was split from the `kano-toolset`
package into its own package. Depend on this package instead.

Required after https://github.com/KanoComputing/kano-toolset/pull/206.


tombettany:
Merging with https://github.com/KanoComputing/kano-toolset/pull/206

PR #380, opened 2016-05-04
Fix sentry startup

Ealdwulf:
This PR fixes a few issues missed in merging #360 : 
* sentry wasn't being restarted at boot except in ultimate parental mode (it is now required in all parental control modes except 'off') 
* opendns wasn't being used (because it now needs to be used at the same time as sentry)
Unrelated to #360, changing between different parental control modes wasn't changing the hosts file, so, eg, going to mode 2 wan't adding search engines to the blacklist if you were already in mode 1, because the loging which wrote the hosts file had a check that assumed that the hosts file couldn't need changing if we it was 'already big'.
Therefore when we are in a parental control mode I have added  code to restore the backup hosts file before appending the blacklist. I think this shoudl be okay (after all, we already restore it when coming out of parental control, so this is effectively like doing that and then going back in). But I don't understand why we didn't do it that way before. Any ideas @skarbat @tombettany ?

There are a number cases here, so I'd like to get it merged soon, so I can ask GAT to test it. 
@alex5imon

### DIFF
--- a/kano_settings/system/advanced.py
+++ b/kano_settings/system/advanced.py
@@ -342,8 +342,10 @@ def set_ultimate_parental(ultimate, safesearch):
         restore_dns_interfaces()
         redirect_traffic_to_google()
         if ultimate:
+            set_setting("use_sentry", "whitelist")
             parse_whitelist_to_config_file(sentry_config)
         elif safesearch:
+            set_setting("use_sentry", "safesearch")
             make_safesearch_config_file(sentry_config)
 
         # Now set resolv.conf to point to localhost
@@ -351,10 +353,13 @@ def set_ultimate_parental(ultimate, safesearch):
         redirect_traffic_to_localhost()
         launch_sentry_server(sentry_config)
 
+
     else:
         restore_dns_interfaces()
         redirect_traffic_to_google()
         kill_server()
+
+        set_setting("use_sentry", "")
 
 
 def redirect_traffic_to_google():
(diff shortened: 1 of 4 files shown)


kano_settings/system/advanced.py
```diff
@@ -282,6 +285,10 @@ def set_hosts_blacklist(enable, block_search,
     if enable:
         logger.debug('enabling blacklist')
 
+        if os.path.exists(hosts_file_backup):
+            logger.debug('restoring original backup file')
+            shutil.copy(hosts_file_backup, hosts_file)
```
tombettany: If we do this and the backup file is large then we end up copying this file and then copying it back again. Also here we do `shutil.copy()` and a few lines down we do `shutil.copyfile`; is there a difference? Should one be preferred?

Ealdwulf: The backup file should be small, as it doesn't contain the parental control stuff.
Good catch re copy/copyfile: Apparently copy propagates the permissions, copyfile does not.  I think we should probably use copy when making the backup as well as when restoring it, so the permissions are propagated both ways, I'll add a change to do that....

kano_settings/system/advanced.py
```diff
@@ -342,32 +349,34 @@ def set_ultimate_parental(ultimate, safesearch):
         restore_dns_interfaces()
         redirect_traffic_to_google()
         if ultimate:
+            set_setting("use_sentry", "whitelist")
             parse_whitelist_to_config_file(sentry_config)
         elif safesearch:
```
tombettany: Are `ultimate` and `safesearch` mutually exclusive? If not then this should be an `if` rather than `elif`.

tombettany:
Otherwise I think that everything is fine here. 👍

kano_settings/system/advanced.py
Ealdwulf: They are mutually exclusive (you can't access search engines in ultimate mode anyway).

PR #381, opened 2016-05-05
Migrating 2 system boot services into systemd

skarbat:
This changeset migrates the `/etc/rc.local` family of scripts into systemd.

 * Core settings script (cpu mode, kano keyboard hotkeys)
 * Kano connect service to get wireless connection on boot

The kano-connect service is not finished yet, as it sometimes fails through systemd.


skarbat:
@Ealdwulf I am discarding this PR as discussed previously. Not applicable at this time.

PR #382, opened 2016-05-09
Switch to pulling whitelist from the current branch.

Ealdwulf:
We are updating the parental control whitelist from github, but from the wheezy branch.
If we every merge back to master, this will need to be update again.
@tombettany 
cloudinary.com has been added to the jessie branch of the whitelist, which may be causing some problem. Any idea what it is used for @pazdera @convolu ?

### DIFF
--- a/kano_settings/system/advanced.py
+++ b/kano_settings/system/advanced.py
@@ -436,7 +436,7 @@ def get_whitelist():
     try:
         online_whitelist = (
             "https://raw.githubusercontent.com/KanoComputing/kano-settings/"
-            "master/WHITELIST"
+            "jessie/WHITELIST"
         )
         html = urllib2.urlopen(online_whitelist).read()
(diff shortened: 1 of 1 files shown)


convolu:
@Ealdwulf I think we were considering using a service for automatic image formatting and cloudinary seems to be one. I don't think we went ahead to use it after all.

Maybe @rcocetta may know whether we use it for our stuff.

convolu:
In fact, @pazdera added it in https://github.com/KanoComputing/kano-settings/commit/51efdcaf0c32a4a03cd37e3878b3ef165c88218e . I don't think we ended up using it, but we may in the future

Ealdwulf:
Thanks @convolu. If we're not using it then the fact that we are using the whitelist from master instead of jessie won't cause a problem yet. I think I'll still merge this PR though.

Ealdwulf:
Merging this because it's a trivial change...

PR #384, opened 2016-05-25
Wallpaper: Hide the option in dashboard mode

tombettany:
Setting the wallpaper only works in desktop mode so don't show the
option in dashboard mode.

cc @skarbat


skarbat:
very true @tombettany

PR #385, opened 2016-06-02
Pi3 fix for uart

Ealdwulf:
Add changes to config for make-light on the pi3, and comments in methods which are now used in the updater.
For https://github.com/KanoComputing/peldins/issues/2369
@alex5imon @tombettany This is the equivalent change needed for the image build.

### DIFF
--- a/boot_default/config.txt
+++ b/boot_default/config.txt
@@ -76,3 +76,4 @@ dtparam=audio=on
 [pi3]
 # for light board
 enable_uart=1
+[all]
--- a/boot_default/config.txt
+++ b/boot_default/config.txt
@@ -73,3 +73,6 @@ dtparam=audio=on
 
 # for more options see http://elinux.org/RPi_config.txt
 
+[pi3]
+# for light board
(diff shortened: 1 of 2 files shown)


boot_default/config.txt
```diff
@@ -73,3 +73,6 @@ dtparam=audio=on
 
 # for more options see http://elinux.org/RPi_config.txt
 
+[pi3]
+# for light board
+enable_uart=1
```
tombettany: My only concern is that potentially we might append to this file, in which case the appended statements would only apply to RPi 3. Maybe it would be wise to finalise with a `[all]`?

More info at the bottom of the [config.txt docs](https://www.raspberrypi.org/documentation/configuration/config-txt.md).

tombettany: I suppose that this comment also applies to https://github.com/KanoComputing/kano-updater/pull/185

Ealdwulf: Good idea, I'll add it to both.

PR #387, opened 2016-06-03
Use alsa-store instead of the deprecated alsa-utils.

Ealdwulf:
This PR updates kano-settings to use alsa-store instead of the deprecated alsa-utils.
@radujipa 
kano-boot-up sound is also changed to depend on alsa-restore instead of kano-utils.


alex5imon:
👍 

@FMog big one for you to test :)

radujipa:
:shipit:

PR #389, opened 2016-06-09
Added permissions config to enable USB drive auto-mount

radujipa:
Related to https://github.com/KanoComputing/peldins/issues/2378

cc @skarbat @Ealdwulf

### DIFF
--- a/debian/kano-settings.install
+++ b/debian/kano-settings.install
@@ -21,7 +21,7 @@ etc/rc.network /etc
 etc/rc.overclock /etc
 etc/rc.others /etc
 etc/init.d /etc
-etc/polkit-1 /etc
+etc/polkit-1/localauthority/50-local.d/* /etc/polkit-1/localauthority/50-local.d
 
 etc_override /usr/share/kano-settings
 
--- a/debian/kano-settings.install
+++ b/debian/kano-settings.install
@@ -21,6 +21,7 @@ etc/rc.network /etc
 etc/rc.overclock /etc
 etc/rc.others /etc
 etc/init.d /etc
+etc/polkit-1 /etc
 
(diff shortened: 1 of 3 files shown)


debian/kano-settings.install
```diff
@@ -21,6 +21,7 @@ etc/rc.network /etc
 etc/rc.overclock /etc
 etc/rc.others /etc
 etc/init.d /etc
+etc/polkit-1 /etc
```
skarbat: @radujipa I think we need to copy the file explicitly, because this directory belongs to a different package, and apt would mistakenly try to clean it up on removal, as in here:

 * https://github.com/KanoComputing/os-dashboard/blob/master/debian/kano-dashboard-common.install#L14

skarbat:
Apart from comment above, looks great to me!

debian/kano-settings.install
tombettany: Yeah, I think that you are right @skarbat.

radujipa: Wow, that's a great catch guys! Thank you!

radujipa:
@skarbat @tombettany Does that fullpath look alright now?

skarbat:
looks good to me @radujipa . Should we add a dependency to `polkit` to make sure it will be there?

radujipa:
@skarbat I think that's a good safeguard to have, but I have a feeling that if that goes missing a lot more things will break than just auto-mounting. I've added the dependency on `policykit-1` for this.

tombettany:
👍

skarbat:
Agreed @radujipa !

radujipa:
Thanks for your help here guys!

PR #390, opened 2016-06-09
Don't change the config if the pi is not recognised

Ealdwulf:
A pi in the workshop was seen to get confused about restoring config.txt because it did not have the fix for https://github.com/KanoComputing/peldins/issues/2366. This PR makes this path more robust in the event that future pis have more revision numbers.
@tombettany @skarbat   
Not necessary to merge for 3.3


skarbat:
👍

tombettany:
👍

Ealdwulf:
After discussion, 2ith @tombettany, decided to merge for 3.3

PR #391, opened 2016-08-08
Support for building spanish image with keyboard settings.

Ealdwulf:
This PR implements the missing the `set keyboard  --layout` option in kano-settings cli, which is used in the es_AR overnight image build.
It makes set_keyboard save the keyboard setting in /etc/default/keyboard, and also makes it set the console setting.

@skarbat @tombettany


skarbat:
I think it looks fine @Ealdwulf

Ealdwulf:
Thanks @skarbat , merging...

PR #392, opened 2016-08-12
Tests: Add "require_rpi" flag

tombettany:
When running the tests, there are some which can't be run if the correct
hardware isn't present. Accommodate this by providing a `require_rpi`
decorator for tests.


tombettany:
This is safe as it is only providing for the tests. Will merge straight away.

skarbat:
I like it @tombettany !

PR #393, opened 2016-08-12
Keyboard locale: Allow setting the keyboard from the CLI

tombettany:
https://trello.com/c/B07XgJVB/201-get-kano-settings-cli-to-implement-setting-keybaord-layout

The CLI is used to configure the locale settings for the image at build
time. As such, expose the keyboard setting interface and unify it with
the GUI setting.

cc @Ealdwulf


Ealdwulf:
Looks nice @tombettany 👍

PR #394, opened 2016-08-15
Boot config filters

tombettany:
https://trello.com/c/z239QJMg/154-5-configure-the-config-txt-to-allow-the-screen-kit-to-be-the-right-way-up-by-default

* Adds support for filters in the boot config file
* Adds display rotation for the screen kit by default
* Some tests for the added functionality

cc @Ealdwulf @skarbat


kano_settings/boot_config_line.py
```diff
@@ -0,0 +1,86 @@
+import re
```
skarbat: Perhaps a short explanation on what this module knows to do?

skarbat:
Looks ok to me @tombettany

tombettany:
@Ealdwulf Right, now all the tests pass :-)

Now to check no EDID, the noobs setting and ensuring that the filtered values always appear at the bottom of the config so they are not overridden.

tombettany:
@Ealdwulf I made sure that there is a test case for the missing EDID and the noobs setting passes from the original tests. There isn't a method to ensure that settings don't get overwritten by later, unfiltered ones but this presents quite a challenge. I think that we should get this merged in so we can begin actual testing of it.

tombettany:
@Ealdwulf Bump. I want to try to get this in as early as possible so that we have time to test it.

Ealdwulf:
Looks okay to me :shipit:

PR #395, opened 2016-08-17
Move pycountry import into function definition

jkimbo:


Ealdwulf:
Thanks. Assuming you tried importing it, it looks good to merge 👍

PR #396, opened 2016-08-24
Locale: Scope XSESSION_RC_FILE to utilising function

tombettany:
The `XSESSION_RC_FILE` file utilises the `get_user_unsudoed()` function
which requires a user to have already been created. This fails on the
build server chroot where no users exist so scope it inside its
utilising function.

Moreover, migrate away from utilising this method for setting the
locale.

cc @Ealdwulf


Ealdwulf:
:ship: :it:

PR #398, opened 2016-09-08
Keyboard setup fixes

jkimbo:
Fixes https://trello.com/c/Zf6k4Oy9/73-kano-init-font-is-too-smal

After quite a lot of investigation I found out the reason as to why the console font on the Spanish image is tiny is because, when [setting up the keyboard layout]( https://github.com/KanoComputing/peldins/blob/69ee1718fe5d072c7f158ab06bc97ee20767fb2c/Makefile#L546), `kano_settings` sets the `XKBLAYOUT` in `/etc/default/keyboard` to `es_AR` which is not recognised and so the console setup fails. The fix was to set `XKBLAYOUT` to just `es` which works fine.

I also fixed a bug in the keyboard settings page that prevented it from displaying on the english image.

cc. @Ealdwulf @radujipa

### DIFF
--- a/kano_settings/set_keyboard.py
+++ b/kano_settings/set_keyboard.py
@@ -13,7 +13,7 @@ import kano_settings.system.keyboard_layouts as keyboard_layouts
 import kano_settings.system.keyboard_config as keyboard_config
 import kano_settings.common as common
 from kano_settings.templates import Template
-from kano_settings.config_file import get_setting, set_setting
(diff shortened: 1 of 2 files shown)


tombettany:
[[Init] Fresh es_AR image has a very small font size in the console on first boot](https://trello.com/c/2nISJ9rO/231-init-fresh-es-ar-image-has-a-very-small-font-size-in-the-console-on-first-boot)

PR #399, opened 2016-09-08
Fix set to uncomment setting

Ealdwulf:
This PR fixes set_config_value to make sure the setting is uncommented if you call "set_config_value()",
unless the value it is passed is None, in which case it comments out the setting.
* set_config_value() should replace commented out version of the setting.
* Had to extend the representation class (boot_config_line) a bit, to clearly distinguish commented out settings and random comments. Commented out settings always need a '=0'.
Updated the unit test to include commented out settings.
@skarbat @radujipa @tombettany

### DIFF
--- a/kano_settings/boot_config.py
+++ b/kano_settings/boot_config.py
@@ -125,7 +125,7 @@ class BootConfig:
         logger.info('writing value to {} {} {}'.format(self.path, name, value))
 
         config = BootConfigParser(lines)
-        config.set(name, value if value else '0', config_filter=config_filter)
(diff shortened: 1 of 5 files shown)


Ealdwulf:
This is for https://trello.com/c/FAp1US5b @rcocetta

skarbat:
A little difficult to follow for me, but good that we keep tests with it @Ealdwulf . 👍

Ealdwulf:
I'm going to merge this for testing, but I'll probably do some more testing on it myself next week.

PR #401, opened 2016-09-13
Terminate filters

Ealdwulf:
This PR should fix https://trello.com/c/cDOFycaq
Here, I only try to make sure all filters are in the form the module understands. I will put in another ticket to enhance it to understand filters properly.
@radujipa @rcocetta @tombettany


radujipa:
Let's test this to the bone! :shipit:

PR #406, opened 2016-09-30
Add sync

Ealdwulf:
This PR adds a sync after the logs are copied to `/boot` in case the customer has to pull the plug to collect the logs.
@tombettany @skarbat


skarbat:
Looks good to me @Ealdwulf .

Just a consideration to keep in mind, I think, is that the script is launched in the background from `kano-safeboot-mode`, when there's probably a lot of R/W into disk going on.

I had a thought recently at wether it would make sense to remount the `/boot` partition, hence forcing a flush into this one alone, but I haven't tried it.

👍

skarbat:
Wonderful @Ealdwulf !

PR #407, opened 2016-10-06
Add boot check

Ealdwulf:
This PR adds a program `kano-boot-check` which tests whether the plug was pulled.
It is run at bootup with the argument `start` and shutdown with the argument `stop`. 
To determine if the system was shut down properly, look at the return code of `kano-boot-chek test`
which is successful (0) if it was.
@skarbat @radujipa 
This can be used to implement the warning which is planned for the dashboard.
For https://trello.com/c/UiqoXEkb

### DIFF
--- a/bin/kano-boot-check
+++ b/bin/kano-boot-check
@@ -6,7 +6,14 @@
 #
 # Check whether the kit has been shut down safely, or the plug has been pulled.
 
-# FIXME add copyright
+# Usage:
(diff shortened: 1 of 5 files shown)


skarbat:
Beautiful @Ealdwulf ! I like the decoupling of the tool from the init.d service.

Ealdwulf:
Thanks @skarbat ! Merging...

PR #410, opened 2016-10-17
Fix spurious shutdown warnings on forced reboot

Ealdwulf:
Sometimes we call reboot without taking down systemd units, which prevents us from
correctly deciding whether the plug has been pulled. Always shut down kano-boot-check.service.

@tombettany @skarbat


skarbat:
I think it's because we use `--force`, grepped and can't find any isolated one.
looks good to me @Ealdwulf .

PR #411, opened 2016-11-07
Test 256mb GPU mem

Ealdwulf:
Change gpu_mem to 256 for test purposes.
Making the change on the RC for 2 reasons: we want to QA it, and this will also test the rc rebase detection . @tombettany


Ealdwulf:
As discussed @tombettany , merging for testing tomorrow.

PR #412, opened 2016-12-06
Fixed screen flip option not being initialised properly

radujipa:
cc @tombettany


radujipa:
[[Settings] Flip screen option isn't being correctly read which causes the screen kit (which is inverted by default) to not function correctly](https://trello.com/c/MMtulvqS/511-settings-flip-screen-option-isn-t-being-correctly-read-which-causes-the-screen-kit-which-is-inverted-by-default-to-not-function-)

PR #414, opened 2016-12-13
Keep minecraft local hostname accurate

skarbat:


Ealdwulf:
Thanks @skarbat. Do you think the version number should change so make-minecraft triggers the isntallation of this version?

PR #418, opened 2017-02-14
i18n: Update es-AR translations

tombettany:
Update the `es-AR` translations.

cc @skarbat


po/es_AR.po
```diff
@@ -951,138 +682,488 @@ msgstr "Escoge tu fondo"
...
+#~ msgid "Unpairing {}"
+#~ msgstr "Desemparejando {}"
+
+#~ msgid "Pairing {}"
+#~ msgstr "Emparejando {}"
 
-msgid "Get ready... Kano is starting up.\n"
-msgstr "Prepárate... Kano está empezando.\n"
+#~ msgid "Get ready... Kano is starting up.\n"
+#~ msgstr "Prepárate... Kano está empezando.\n"
```
tombettany: @skarbat Did we manually add this one last time?

tombettany:
[Settings](https://trello.com/c/gKG9ZhOp/16-settings)

skarbat:
This is correct @tombettany - I did that because we needed a way to pull the translation from a script:

 * https://github.com/KanoComputing/kano-settings/blob/master/etc/init.d/kano-settings#L17

Ealdwulf:
That can be automated like this:
https://github.com/KanoComputing/make-snake/blob/master/po/Makefile#L27, I
think.

skarbat:
@tombettany, Just pushed 3 small translation typos - Looks good to me.

tombettany:
@Ealdwulf I made your suggested changes. Merging now for testing.

etc/init.d/kano-settings
```diff
@@ -10,11 +10,13 @@
 ### END INIT INFO
 
 . /lib/lsb/init-functions
+. gettext.sh
```
skarbat: should this have an absolute path, given it's pulled from a former sysv init script?

tombettany: Hmm, we'll try it and see what happens...

PR #427, opened 2017-03-15
Do not blink the board LED unless specified.

skarbat:
@tombettany


skarbat:
[(5) As an engineer, it should be clear what is started at startup and from where it is started](https://trello.com/c/BE8qaRqf/787-5-as-an-engineer-it-should-be-clear-what-is-started-at-startup-and-from-where-it-is-started)

PR #428, opened 2017-03-16
Add support for lower&faster resolution defaults

skarbat:
This changeset is to provide an automated process to choose a lower screen resolution by default.


skarbat:
[(5) As a user, I want my kit to operate at a lower resolution by default so that it runs smoother [3]](https://trello.com/c/FctxWc5p/763-5-as-a-user-i-want-my-kit-to-operate-at-a-lower-resolution-by-default-so-that-it-runs-smoother-3)

skarbat:
This feature is starting to work at my end. 
We'd need to expand the tests with more screen dumps until we're happy with it.

@tombettany @Ealdwulf

tests/optimal_screen/screen-adafruit-hdmi.dump
```diff
@@ -0,0 +1,89 @@
+{
```
tombettany: Is it possible to instead use the dump output from `tvservice --dumpedid <filename>` or the output from `edidparser <monitor_dump>` rather than an already semi-parsed version? It would make reaming this information from monitors easier and allow us to test the parsing step as well.

bin/kano-settings-onboot
```diff
@@ -30,35 +58,28 @@ from kano_settings.system.audio import is_HDMI, set_to_HDMI
...
 def check_model_present(model):
+    '''
+    Returns True if the requested screen model is active
+    '''
     if get_config_comment('kano_screen_used', model):
-        logger.info('The unit was configured for this model already, exiting.')
-        sys.exit()
+        return True
+    else:
+        return False
```
tombettany: Just

```python
return get_config_comment('kano_screen_used', model) not None
```

or something like that?

bin/kano-settings-onboot
```diff
@@ -30,35 +58,28 @@ from kano_settings.system.audio import is_HDMI, set_to_HDMI
...
 
 def override_models():
-    global overrides, edid, model
+    '''
+    This function sets specific modes for screens that require special treatment
+    '''
+    overrides = {
+        '32V3H-H6A':     { 'target_group': 'DMT', 'target_mode': 16, 'is_monitor': True },
+        'AS4637_______': { 'target_group': 'DMT', 'target_mode': 16, 'is_monitor': True },
+        'BMD_HDMI':      { 'target_group': 'CEA', 'target_mode': 33, 'is_monitor': True },
```
tombettany: Maybe add a comment to each of these screens to describe what they are (first 2 I think are the screen kit)

bin/kano-settings-onboot
```diff
@@ -135,27 +158,122 @@ def compare_and_set_overscan():
...
+    logger.info('Searching for the best optimal resolution')
+
+    def find_best_resolution_aspect(aspect, width, height):
+        '''
+        Finds the screen mode at the given resolution and aspect ratio,
+        If aspect is None, the first matching resolution mode will be returned, if found.
+        '''
+        for mode in supported:
+            w = int(mode['resolution'].split('x')[0])
+            h = int(mode['resolution'].split('x')[1])
```
tombettany: Maybe abstract this to the parser so that we have something along the lines of

```python
mode['resolution'] = {
    'width': <resolution.split('x')[0]>,
    'height': <resolution.split('x')[1]>
}
```

bin/kano-settings-onboot
```diff
@@ -135,27 +158,122 @@ def compare_and_set_overscan():
...
+            if w == width and h == height:
+
+                # If we care about aspect ratio, skip it if no match
+                if aspect and aspect != mode['aspect']:
+                    continue
+
+                logger.info('Found best mode at {}x{} mode: {} aspect: {}'.format(
+                        width, height, mode['mode'], mode['aspect']))
+                return mode
```
tombettany: Are these modes pre-sorted?

bin/kano-settings-onboot
```diff
@@ -164,132 +282,174 @@ def ensure_correct_browser():
...
+        # write comment to config
+        set_config_comment('kano_screen_used', model)
+        end_config_transaction()
 
-# changes
-changes = compare_and_set_mode() or compare_and_set_full_range() or \
-    compare_and_set_overscan()
+        # reboot
+        if dry_run:
+            logger.debug("dry_run rebooting due to config changes")
```
tombettany: Maybe `dry_run` should output the `config.txt` file that it would write since it isn't actually writing it.

PR #429, opened 2017-03-30
Fixing test for optimal screen

skarbat:
@radujipa


radujipa:
Happy with this if it works :shipit: However, I think the `=False` will happen before the `getenv` 🤔

PR #430, opened 2017-03-30
Add an API to detect the Kano Screen Kit

skarbat:
Needed by PR: https://github.com/KanoComputing/kano-desktop/pull/226


kano_settings/system/display.py
```diff
@@ -263,6 +263,13 @@ def is_mode_fallback():
     return w == 640 and h == 480
 
 
+def is_screen_kit():
+    '''
+    Returns True if the screen is a Kano Screen Kit.
+    '''
+    return get_model() in ('ADA-HDMI', 'MST-HDMI1')
```
radujipa: Looking at how `get_model` is implemented, I think it's quite odd. Could you please use `get_edid_name(use_cached=False)` ?

skarbat: done.

radujipa:
:shipit:

PR #431, opened 2017-04-07
Signal poweroff to a GPIO pin

skarbat:
This changeset will set GPIO pin 26 high when the kit is powered off from Dashboard, Desktop or command line (sudo poweroff / halt, etc). It has no effect during a reboot.

This is what will happen:

 * System will terminate all apps and do a normal shutdown
 * The init process will flush buffers then set the GPIO high
 * Init will wait for 3 seconds for the power to be unplugged
 * Init will exit gracefully
 * A harmless kernel panic will be displayed if this step es reached, this is not considered an error.

It is possible to reverse the gpio state with `active_low=y`: https://github.com/raspberrypi/firmware/blob/master/boot/overlays/README

@tombettany @radujipa


Ealdwulf:
Have you checked that the ck-lite power-on mechanism still works when you do this? 
See https://www.raspberrypi.org/forums/viewtopic.php?p=783074&sid=52f1d52fd0cf2b3ad2141513d1b3dcbe#p783074 
" Another feature that cannot be used anymore is to reboot the Pi from the halted state by shortening SCL0 or GPIO-4 to GND. Using P6 is now the only method, apart from pulling the plug."

skarbat:
It certainly stops functioning. But I think the purpose of this feature is to cut the power off the board anyway.

skarbat:
I am closing this PR, not applicable anymore.

PR #433, opened 2017-04-26
Added PiHat to user preference when enabling the LED ring anim

radujipa:
There is an option in `Notifications` to turn ON/OFF the cpu-monitor animation on the LED ring. This adds support for the PiHat as well.

@tombettany @skarbat


skarbat:
👍

PR #435, opened 2017-05-04
Fixed a number of issues with automatic lower resolution setup

radujipa:
@skarbat @tombettany


radujipa:
[(5) As a user, I want my kit to operate at a lower resolution by default so that it runs smoother [8]](https://trello.com/c/FctxWc5p/763-5-as-a-user-i-want-my-kit-to-operate-at-a-lower-resolution-by-default-so-that-it-runs-smoother-8)

kano_settings/system/display.py
```diff
@@ -185,7 +185,7 @@ def set_hdmi_mode(group=None, mode=None):
     group = group.lower()
     mode = int(mode)
 
-    if group == 'cea':
+    if group == 'cea' or group == '1':
```
tombettany: Can we also include the non-string `1`? It seems that it is likely that you would want to actually call this with integers as this is how you presumed it was behaving.

In which case, we might want to clean up the line with something like

```python
if group in ['cea', '1', 1]:
```

Also, do we trust that `mode` will always be an integer or string representation of an integer, otherwise `mode = int(mode)` above will fail.

Finally, can you add a comment to this function - it may not be a great function but at least having some explanation about what it actually does rather than what we assume it does would be helpful.

bin/kano-settings-onboot
```diff
@@ -429,13 +491,21 @@ if __name__ == '__main__':
         set_to_HDMI(False)
 
     # calculate changes needed
-    changes = compare_and_set_mode() or \
-        compare_and_set_full_range() or \
-        compare_and_set_overscan() or \
-        compare_and_set_optimal_resolution()
+    overrall_changes = False
```
tombettany: This variable is misspelt - should be `overall_changes`. Not too much of a problem because it is defined and anyone reading this would understand what it means but can be a pain for anyone wishing to use the variable later.

bin/kano-settings-onboot
```diff
@@ -429,13 +491,21 @@ if __name__ == '__main__':
...
+    changes = compare_and_set_full_range()
+    overrall_changes = overrall_changes or changes
 
+    changes = compare_and_set_overscan()
+    overrall_changes = overrall_changes or changes
+
+    changes = set_resolution_for_screen_kit() or compare_and_set_optimal_resolution()
+    overrall_changes = overrall_changes or changes
+
+    if changes:
```
tombettany: Should this be `overall_changes`? If so, presumably the above lines can be reduced to

```python
overall_changes = \
    compare_and_set_mode() \
    or compare_and_set_full_range()
    or changes = compare_and_set_overscan()
    or set_resolution_for_screen_kit() \
    or compare_and_set_optimal_resolution()
```

Hmm, except if we actually require many of them to make changes in which case it might stop on the first call returning `True`. I don't know what behaviour you actually want but either way it seems like you can completely remove the `changes` (or `overall_changes`) variable and simplify above.

bin/kano-settings-onboot
```diff
@@ -389,23 +449,25 @@ if __name__ == '__main__':
...
-
     supported = screen_data['supported']
     if not supported:
         logger.debug("could not determine list of screen supported modes, aborting")
         sys.exit(1)
 
+    edid = screen_data['edid']
+    if not edid:
+        changes = set_resolution_for_unflashed_screen()
+        if changes:
```
tombettany: Up to you if you want to combine the `if` with the method call.

bin/kano-settings-onboot
```diff
@@ -350,14 +411,14 @@ if __name__ == '__main__':
...
-        set_safeboot_mode()
-        reboot_now = True
+    # # Reconfigure and reboot if the user requested safe mode
+    # # Or if the cable appears not to have been plugged in.
+    # if is_mode_fallback():
+    #     # Backup the config file and reboot into safe mode
+    #     logger.warn("executing fallback boot")
+    #     safe_mode_backup_config()
+    #     set_safeboot_mode()
+    #     reboot_now = True
```
tombettany: Completely remove?

bin/kano-settings-onboot
```diff
@@ -205,7 +267,7 @@ def compare_and_set_optimal_resolution():
             if mode['mode'] == preferred_mode and mode['group'] == preferred_group:
                 preferred_aspect_ratio = mode['aspect']
                 logger.info('Reported preferred mode is {} group {} aspect ratio: {}'.format(
-                        preferred_mode, preferred_group, preferred_aspect_ratio))
+                    preferred_mode, preferred_group, preferred_aspect_ratio))
```
tombettany: Some weird stuff with parentheses matching making it difficult to follow. Probably should newline the matching `)`s, maybe something like

```python
logger.info(
    'Reported preferred mode is {} group {} aspect ratio: {}'.format(
        preferred_mode, preferred_group, preferred_aspect_ratio
    )
)
```
or
```python
logger.info(
    'Reported preferred mode is {} group {} aspect ratio: {}'
    .format(preferred_mode, preferred_group, preferred_aspect_ratio)
)
```

bin/kano-settings-onboot
```diff
@@ -158,6 +159,87 @@ def compare_and_set_overscan():
...
+    set_hdmi_mode(optimal_mode['group'], optimal_mode['mode'])
+    return True
+
+
+def find_best_resolution_aspect(aspect, width, height):
+    '''
+    Finds the screen mode at the given resolution and aspect ratio,
+    If aspect is None, the first matching resolution mode will be returned, if found.
+    '''
+    global supported
```
tombettany: Can we pass this instead of `global`ing it?

bin/kano-settings-onboot
```diff
@@ -158,6 +159,87 @@ def compare_and_set_overscan():
...
+
+def set_resolution_for_unflashed_screen():
+    '''
+    This function sets a predefined resolution for a display
+    without EDID information.
+
+    Returns:
+        changes - bool whether or not there were changes made
+    '''
+    global status
```
tombettany: Again, are we able to pass `status`?

radujipa:
@tombettany Fixed the issues you pointed out.

bin/kano-settings-onboot
```diff
@@ -429,13 +472,22 @@ if __name__ == '__main__':
...
+    overall_changes = overall_changes or changes
 
+    changes = compare_and_set_overscan(edid, status)
+    overall_changes = overall_changes or changes
+
+    changes = set_resolution_for_screen_kit(supported) or \
+              compare_and_set_optimal_resolution(edid, status, supported, verbose)
+    overall_changes = overall_changes or changes
+
+    if changes:
```
tombettany: Thanks for the changes Radu. This is the only line that I'm still concerned about - should this `changes` only be the last version of the values that you check or should it be `overall_changes`? Obviously you keep overwriting `changes` and I just want to check that this is what you intend.

tombettany: If it is as intended then I'm happy to go ahead and merge but just double checking.

radujipa: No, you're absolutely right.. Obviously that's what those lines were working up towards..

Good spot, thanks 👍

radujipa:
@tombettany This is ready to be merged on my side. Any other thoughts here?

PR #436, opened 2017-05-08
Ability to set HDMI audio with --force (when EDID is missing)

radujipa:


radujipa:
[When a video plays and PiHat is connected, audio is forced though the jack causing the LEDs to light up frantically](https://trello.com/c/hzyr5RTS/872-when-a-video-plays-and-pihat-is-connected-audio-is-forced-though-the-jack-causing-the-leds-to-light-up-frantically)

PR #437, opened 2017-05-09
Fix test_adafruit_hdmi test case

radujipa:
The test case was calling kano-settings-onboot from a relative path
to the test script, but was being run one level up. Prepended the
relative paths with the fullpath to the script to point to the
files correctly.

Noticed this in the daily QA tests report.


tests/optimal_screen/test_optimal_screens.py
```diff
@@ -18,14 +18,15 @@
...
 
 # Import kano-settings from this cloned repository
-import sys
 sys.path.insert(1, '../../')
 
-verbose=False
-app='sudo ../../bin/kano-settings-onboot'
+verbose = False
+script_path = os.path.dirname(os.path.realpath(__file__))
+app = 'sudo ' + os.path.join(script_path, '..', '..', 'bin', 'kano-settings-onboot')
```
skarbat: Is it possible to use absolute paths? Just in case we need testing it from elsewhere...

radujipa: Good point @skarbat 👍  

They were already fullpaths, but it's best to also have them absolute without `..` inside.

PR #438, opened 2017-05-22
Option to skip confirmation screen when there is internet via ether

radujipa:
The need for this option comes from the OS onboarding flow where we
launch `kano-wifi-gui` to encourage users to connect to the internet.
However, if they've already connected via an ethernet cable there is
no need to ask them to connect via WiFi at that point.


radujipa:
[(P4) Onboarding - Connect to WiFi comes up despite already being connected via ethernet](https://trello.com/c/Ay0XXsjN/1069-p4-onboarding-connect-to-wifi-comes-up-despite-already-being-connected-via-ethernet)

radujipa:
[Onboarding - WiFi setup comes up even though already connected via ethernet](https://trello.com/c/JsqqxoNN/1208-onboarding-wifi-setup-comes-up-even-though-already-connected-via-ethernet)

PR #439, opened 2017-05-22
Fixing Youtube not playing with low parental control

skarbat:
On low parental control, the youtube videos stopped playing.
Reason being that the `forcesafesearch` filter was being applied which is no longer appropiate.
Fixed by wrapping official youtube-neede URLs to redirect to a restricted youtube site.

@tombettany @radujipa


skarbat:
[(P2) YouTube - If "Low" Parental Lock is enabled, YouTube videos don't play](https://trello.com/c/VYNutw1e/875-p2-youtube-if-low-parental-lock-is-enabled-youtube-videos-don-t-play)

kano_settings/system/advanced.py
```diff
@@ -415,13 +415,19 @@ def make_safesearch_config_file(config_file, opendns):
     else:
         # use google servers 
         servers = ', '.join(google_servers)
+
     import pycountry, json
-    rule = "cname ^(?!forcesafesearch)(.*).(?:google|youtube).(?:{}) to forcesafesearch.google.com using "+servers    
+
+    rule_google = "cname ^(?:www.google.)(?:{}) to forcesafesearch.google.com using " + google_servers
+    rule_youtube = "cname ^(?:www.youtube.|m.youtube.|youtubei.googleapis.|youtube.googleapis.|www.youtube-nocookie.)' \
```
tombettany: A little confused here - you start the string with a `"` and end it with a `'` only for it to continue on the next line with a `'` and end with a `"`. Methinks that this isn't quite right. Also if you are going to break the line, keep the following `(?:{})` with it so that the urls aren't broken up. Something like

```python
rule_youtube = "cname ^(?:www.youtube.|m.youtube.|youtubei.googleapis.|youtube.googleapis.|www.youtube-nocookie.)(?:{}) " \
    "to restrict.youtube.com using " + google_servers
```

skarbat:
Thanks guys, found additional typo too. Tested on the PI and works great now - Merging.

PR #440, opened 2017-05-23
Fix automatic resolution setting at first boot

radujipa:
The cause for this was that `kano-settings-onboot` detected a value
for `hdmi_mode` and `group` as being set. However, these lines were
commented out, breaking the assumption that it was already configured. 
These changes allow us to get a config value ignoring commented lines.


radujipa:
[(P2) Screen resolution sometimes set to 1920x1080px](https://trello.com/c/jUoPr5Aq/965-p2-screen-resolution-sometimes-set-to-1920x1080px)

PR #442, opened 2017-05-24
Fix Audio screen to show available features and system status

radujipa:
The change came from having to disable jack audio because of PiHat.
The radio buttons get disabled if audio through HDMI or JACK are
not supported. Also, now they show what is currently set on the
system. **One risk of disabling the HDMI audio button** is that EDID
info may be full of lies, in which case users might need to force
the setting.


radujipa:
[(2) (P2) CK2 Lite: Configuring JACK audio setting should be disabled in Settings when the PiHat is attached](https://trello.com/c/BgxBhJor/986-2-p2-ck2-lite-configuring-jack-audio-setting-should-be-disabled-in-settings-when-the-pihat-is-attached)

PR #443, opened 2017-05-25
Speed up boot time into the Dashboard

skarbat:
This changeset decreases the time it takes to bootup and reach the Dashboard by about 15 seconds on RaspberryPI 3. It does so by disabling the service `console-setup` which takes about 35% of boot time.

This service is responsible for setting up the console font and keyboard, but we don't use such apps anymore.

Additionally it sets a larger console font for text messages, very early on boot.
Tested with a regular kit, as well as with a low battery unit boot up messages.

@tombettany


skarbat:
[(P1) Onboarding - Dashboard takes a long time to load](https://trello.com/c/YH0rKYVD/985-p1-onboarding-dashboard-takes-a-long-time-to-load)

PR #444, opened 2017-06-07
Add a display rotation setting for the new Screen Kit

radujipa:
Luckily the new SK has a slightly different EDID tag and is the
correct way up. Technically, this setting is not needed given the
default value in the file already, but we explicitly set it to
know about it.


radujipa:
[(1) As a CK2 Pro user, I want the screen to be the correct orientation when I boot up the first time](https://trello.com/c/FItzeKLU/921-1-as-a-ck2-pro-user-i-want-the-screen-to-be-the-correct-orientation-when-i-boot-up-the-first-time)

PR #446, opened 2017-06-16
Added messages to Settings > Audio on disabled options

radujipa:
Users need to be made aware why they can't select a particular
option. This adds such messages to the only cases known to disable
either HDMI or Jack audio. Unfortunately, the messages were added
to the UI class rather than in the place that's disabling them.


kano_settings/set_audio.py
```diff
@@ -82,6 +111,12 @@ def current_setting(self):
         analogue_supported = is_analogue_audio_supported()
         hdmi_selected = is_HDMI()
 
+        if not hdmi_supported:
+            self.disabled_hdmi_label.get_style_context().add_class('normal_label')
```
radujipa: This line effectively shows the message and it does so by setting the styling of the text from a Gtk default of `white`...

radujipa:
@skarbat Will merge this for us to test quickly in `rc`, but do please still review.

radujipa:
[As a CK Lite user, I want to know why Analogue audio option is disabled in Settings > Audio](https://trello.com/c/LKl897Js/1087-as-a-ck-lite-user-i-want-to-know-why-analogue-audio-option-is-disabled-in-settings-audio)

skarbat:
👍

PR #447, opened 2017-06-19
Increased power output on the USBs for RPi2 only to 1.2A

radujipa:
The hardware allows for increasing the power output from 600mA to
1.2A for RPi1 B+ and RPi2. Given our support for >=RPi2, we change
this for existing users whom might buy a new Pixel Kit. We need to
ensure in testing that this does not create system instabilities!


radujipa:
I am currently testing this myself with a Pi2. We might need to hold from merging as we know more about the Pro battery.

radujipa:
[As a CK2 Pro user, I want devices that use lots of power to work when plugged into my USB hub](https://trello.com/c/cwrMzGuI/1081-as-a-ck2-pro-user-i-want-devices-that-use-lots-of-power-to-work-when-plugged-into-my-usb-hub)

radujipa:
More info here:

http://hackaday.com/2015/04/06/more-power-for-raspberry-pi-usb-ports/
https://www.raspberrypi.org/forums/viewtopic.php?p=594183#p594183
https://www.reddit.com/r/raspberry_pi/comments/4b2enb/psa_raspberry_pi_3_provides_12a_usb_current_by/
https://www.raspberrypi.org/forums/viewtopic.php?f=28&t=105502

radujipa:
Some basic usage testing shows that without this flag and the following setup:

```
RPi2 + WiFi dongle + Kano Keyboard + Optical Mouse + Pixel Kit + Power Supply
```

suffers from USB disconnects. That's to say that the WiFi drops, keyboard and mouse behave like they are being plugged and unplugged repeatedly. This continues until the Pixel Kit is unplugged.

With the flag on, the above doesn't happen and also I have not noticed the power indicator (lighting) come on in the top right of the screen.

radujipa:
@skarbat You are absolutely right. I double checked that and there's no update strategy for the `config.txt` file. Historically, we just edited the file in a postupdate scenario. We did this enough times for there to be a util function for this honestly :)

skarbat:
👍 @radujipa

PR #449, opened 2017-07-13
Disable the bootup sound

skarbat:
Since the introduction of the kano-peripherals daemons, the sound driver is disabled until user space programs come into play. This changeset simply disables the init script, which we should restore when we find a way to enable sound at this point.


skarbat:
[(2) As a user, I want to hear a delightful sound when I open an app from the dashboard](https://trello.com/c/KM7VODXw/1102-2-as-a-user-i-want-to-hear-a-delightful-sound-when-i-open-an-app-from-the-dashboard)

PR #450, opened 2017-07-27
Control Dropbear SSH client preference

radujipa:
## Added user setting to enable / disable SSH client

This adds a checkbox in the Advanced panel to enable or disable the Dropbear SSH client. The functions have been separated such that they can be used externally.

## Mechanism to enable SSH client at boot

The Dropbear SSH client will be disabled by default in release images and enabled otherwise. This allows to easily enable the SSH by looking for the `/boot/ssh` file. This implementation also follows Rasbian's Pixel mechanism.

kano-screenshot-thu-jul-27-18-44-32


radujipa:
[(5) (P3) SSH enabled by default + default password puts kits at risk to be turned into a member of an evil botnet army](https://trello.com/c/XWHy70rg/940-5-p3-ssh-enabled-by-default-default-password-puts-kits-at-risk-to-be-turned-into-a-member-of-an-evil-botnet-army)

PR #453, opened 2017-08-25
Disable temperature and low voltage overlay warnings only

radujipa:
This change comes amid manufacturing IC specifics for the CKC
battery. As such, we disable the Raspbian overlay warning icons
only and keep the safety mechanism still active (avoid_warnings=2
disables safety as well). Throttling still happens on this setting
and is easily testable by using a programmable power supply or
setting the CPU temp_limit=45 option.

Together with: https://github.com/KanoComputing/kano-updater/pull/252


radujipa:
@Ealdwulf I've added the `PostUpdate` scenario as well in the Updater to apply the patch to existing users. We need this as we've already shipped a bunch of CKCs.

radujipa:
[(3) As a CK2 Pro user, I don't want to be warned about low voltage levels all of the time [5]](https://trello.com/c/0SUXjR9C/1214-3-as-a-ck2-pro-user-i-dont-want-to-be-warned-about-low-voltage-levels-all-of-the-time-5)

boot_default/config.txt
```diff
@@ -103,4 +103,7 @@ max_usb_current=1
 [pi3]
 # for light board
 enable_uart=1
+
+# Disable temperature and low voltage overlay warnings only, safely is still ON.
```
tombettany: `safely` again :p

radujipa: Suffering from _Shooting From The Hip Syndrome_ ...

PR #454, opened 2017-08-29
Removed duplication of setting /etc/hosts for Hack Minecraft

radujipa:
Hack Minecraft has a postinst script to add a line to /etc/hosts
file in order to be able to connect to the game local server.
There does not seem to be any reason why this configuration needs
to be duplicated in this project.

**NOTE**: This was my first attempt at fixing the Trello issue attached below. Upon further investigation and testing, this did not fix the issue - but it did clean up the deb package a bit. The actual cause was this PR https://github.com/KanoComputing/peldins/pull/2833


radujipa:
[(3) [P1] Hack Minecraft broken in devel [0.5]](https://trello.com/c/iuGrgXSa/1240-3-p1-hack-minecraft-broken-in-devel-05)

Ealdwulf:
Maybe the idea was to migrate it from hack-minecraft to here, to keep all the networking settings in one repo?

radujipa:
@Ealdwulf As I see it, HM requires just one line in the `/etc/hosts` file, hence it should be the one installing it. The other change it makes requires more knowledge of the system in that it should also propagate the change to the parental controls backup file of `/etc/hosts`, in which case it should also install that (it only needs to know to update the backup because there's no abstraction mechanism to adding stuff to the hosts file, or one that doesn't make a copy of the file and edits it accordingly).

tombettany:
@radujipa Should we merge this one?

radujipa:
@tombettany I've updated the PR, I think we should merge

radujipa:
[As an engineer, I do not want duplicated configuration for Hack Minecraft in Kano Settings](https://trello.com/c/gfNEpfWt/1490-as-an-engineer-i-do-not-want-duplicated-configuration-for-hack-minecraft-in-kano-settings)

tombettany:
Merge away :)

PR #456, opened 2017-09-07
Dhcp migration

skarbat:
Replace the dhcp client package, to upgrade to a more modern and faster version.
Removing the interfaces definition, as it is taken care of already using the defaults.
The new daemon will be detected by the WPA supplicant for wireless networks, and Ethernet too
This migration also should fix the power off / reboot issue taking ages to complete.

@tombettany @radujipa


skarbat:
[(5) As a CKC user, I want my CKC to reconnect to WiFi as soon as possible after losing connection [5]](https://trello.com/c/OsBtNHkG/1336-as-a-ckc-user-i-want-my-ckc-to-reconnect-to-wifi-as-soon-as-possible-after-losing-connection)

debian/kano-connect.postinst
```diff
@@ -26,13 +29,17 @@ case "$1" in
...
 
-            # updating the /etc/network/interfaces
-            cp /usr/share/kano-network/etc/network/* /etc/network/
+            # remove previous interfaces file, for udhcpd5 migration
+            if [ -f $interfaces_file ]; then
+                rm -f $interfaces_file
+            fi
 
+            if [ -f $interfaces_file_backup ]; then
+                rm -f $interfaces_file_backup
```
tombettany: Do these files have anything to do with the way that we set the parental settings. I think not because I think that we actually care about the `/etc/resolvconf/interface-order` but I'm just a little concerned because I don't understand why we have an `interfaces` and an `interfaces-backup` to start with.

debian/changelog
```diff
@@ -1,3 +1,10 @@
+kano-settings (4.0.2-0) unstable; urgency=low
+
+  * Changes for Kano OS 3.12.0:
```
tombettany: This image definitely won't be 3.12.0. It is likely that it will end up in 3.13.

tombettany:
Oh, and it doesn't merge...

debian/kano-connect.postinst
skarbat: The file `interfaces` originally comes from a postinst script in the`ifupdown` Debian package, with an empty stub. For historic reasons, we saved a backup copy of it to be safe, but we never used the backup version. We did for the original file to force the connection hooks, which seemed to be the cause for troubles during a shutdown. It would still be useful to use for special cases like fixed IP addresses.

I have documented it briefly [here](https://github.com/KanoComputing/kano-toolset/pull/254/commits/53a67cd5a5c83c91153f98a3b4ce0e90c40bab8e)

debian/changelog
tombettany: Just change this to 3.13.0

PR #459, opened 2017-09-12
Refactored automatic resolution policy to use preferred group

radujipa:
## Refactored automatic resolution policy to use preferred group 

The resolution policy before ignored the preferred hdmi_group coming from screen EDIDs. As well as taking into account the preferred group, now the kano-settings-onboot script does fewer reboots and has explicit default resolutions set for all our screens. The hdmi_drive option was as such disabled to avoid NO SIGNAL on some monitors.

## Added display configurations for all SKs to avoid reboots

In order to avoid having to detect the optimal resolution for all our SKs and then reboot, we explicitly set the modes in the config file. The default hdmi_group/mode have also been set for an uSK. Lastly, the hdmi_drive=2 option was removed in order to avoid forcing some monitors to output an unsupported signal.

## Moved old tests to pytest with Make and added tests for Display 

This commit is two fold. Firstly, all the old tests were moved under pytest with a standard Makefile to run them. Here, the TestConfig class was disabled because it required sudo permissions. Secondly, tests for the automatic resolution policy were added. These use fixtures to load a dataset of screen dumps and run check that the display functions return & apply the correct settings.

## Added display testing dataset and tool to generate

This dataset consists of three parts. Firstly, the raw EDID data was dumped from various screens. Secondly, a corresponding RPi dumps file is generated from output of tools used within our display functions (tvservice, edidparser). And thirdly, an expected results file derived from the dumps data to provide the correct return values expected from display functions. The script used to generate these files was also added.

--

Requires: https://github.com/KanoComputing/kano-updater/pull/256


radujipa:
[(8) [CARE] Screen Resolution issue when moving from TV to monitor. [8]](https://trello.com/c/vsFlFYaY/1172-8-care-screen-resolution-issue-when-moving-from-tv-to-monitor-8)

radujipa:
@tombettany @skarbat I've finished the work on this PR (~99%) and can be reviewed now. I recommend looking by commit rather than at the whole thing :)

I still have to add more comments and bump the changelog version. Then I'll add a PostUpdate scenario in the Updater to complement.

kano_settings/system/display.py
```diff
@@ -245,7 +300,8 @@ def set_safeboot_mode():
 def get_status():
     status = dict()
 
-    status_str, _, _ = run_cmd(tvservice_path + ' -s')
+    cmd = '{tvserice} --status'.format(tvserice=tvservice_path)
```
tombettany: Typo in `tvserrice`?

kano_settings/system/display.py
```diff
@@ -379,25 +440,30 @@ def read_hdmi_mode():
...
+        if current_mode['mode'] == mode and current_mode['group'] == group:
+            return index + 1
 
     # 0 for auto
     return 0
 
 
 def read_edid():
-    edid_dat_path = '/tmp/edid.dat'
+    delete_file(TMP_EDID_DAT_PATH)
```
tombettany: Would we also like a `use_cached` option here?

kano_settings/system/display.py
```diff
@@ -518,3 +587,182 @@ def set_gfx_driver(enabled):
                 logger.error("Error restoring fpturbo_config", exception=e)
     end_config_transaction()
     set_setting('Use_GLX', enabled)
+
+
+def override_models(edid, model):
```
tombettany: Do we still need any of these functions below?

boot_default/config.txt
```diff
@@ -75,35 +76,57 @@ display_rotate=0
...
 
-# for more options see http://elinux.org/RPi_config.txt
+[pi2]
+# Give more current to the USB ports for the Pixel Kit.
+max_usb_current=1
+[all]
 
-# Configuration Filters
+[pi3]
+# for light board
```
skarbat: I wasn't aware we don't force max_usb_current on the PI3. We are sure we don't need it for the USB hub, right?

bin/kano-settings-onboot
```diff
@@ -1,353 +1,151 @@
...
-
-    return False
-
-
-def set_resolution_for_unflashed_screen(status):
-    """
-    This function sets a predefined resolution for a display
-    without EDID information.
+from kano_settings.return_codes import RC_SUCCESS, RC_REQUIRES_ROOT, RC_INCORRECT_ARGS, \
+    RC_EDID_NO_MODEL, RC_EDID_NO_STATUS, RC_EDID_NO_SUPPORTED, RC_NO_EDID
```
skarbat: Isn't it more practical to just include them all?

bin/kano-settings-onboot
```diff
@@ -360,172 +158,132 @@ def ensure_correct_volume_levels():
...
 
-    changes = compare_and_set_overscan(edid, status)
-    overall_changes = overall_changes or changes
+def reboot(dry_run=False):
+    """
+    Flush logs to disk and reboot the kit.
+    """
+    if not dry_run:
+        logger.sync()
+        run_cmd('kano-checked-reboot changes systemctl reboot --force')
```
skarbat: Are we sure about this one? this `sync` simply flushes the logger buffers, not the filesystem.
Specially delicate with a `--force` reboot...

 * https://github.com/KanoComputing/kano-toolset/blob/master/kano/_logging.py#L214

bin/kano-settings-onboot
```diff
@@ -360,172 +158,132 @@ def ensure_correct_volume_levels():
...
-    if overall_changes:
-        if verbose:
-            print 'changes applied:', changes
+if __name__ == '__main__':
+    logger.force_log_level('info')
 
-        # write comment to config
-        set_config_comment('kano_screen_used', model)
-        end_config_transaction()
+    if os.geteuid() != 0:
```
skarbat: not really that important, but I think `getuid` is safer and clearer in this case.
We won't really use privilege escalation through `setuid` bit no?

kano_settings/system/display.py
radujipa: Totally a typo, thanks for spotting it!

boot_default/config.txt
radujipa: The `max_usb_current` option is only for Pi2, for Pi3s the USBs already push out the max current.

bin/kano-settings-onboot
radujipa: It might be, but the issue is that the IDE I'm using (Sublime Text) then cannot trace where some functions are getting imported from and cannot warn me if some functions have not been imported at all (because of `from x import *`).

kano_settings/system/display.py
radujipa: Yup :)

kano_settings/system/display.py
radujipa: It would probably benefit from it since we don't use HDMI hotplugging currently, but for the purpose of this PR, I didn't want to modify (too much) what we currently had. I think we should do it in another PR.

bin/kano-settings-onboot
radujipa: Reading back on it, it's definitely clearer as you suggest. Thanks for catching that 👍

bin/kano-settings-onboot
radujipa: I see your point here @skarbat. The only reasons I'd be reluctant to change it is 3 things:

1. This preserves the functionality we had before and didn't have problems.
2. The `end_config_transaction` right above the call here already does an `os.system('sync')`.
3. It would slow it down even more - Why do we even need to tell the OS to flush the filesystem in the first place?

Also, as to why we're using `--force` there is beyond me...

skarbat: Oh okay, I wasn't aware of point 2). Your question about `--force` was because requesting a reboot in the middle of a bootup would be asking for troubles.

PR #461, opened 2017-09-26
Volume boost ckc v1

radujipa:
## Added Volume Boost option for CKCv1

This option in Settings > Audio comes from CKCv1 speaker mitigation. Here, we give users with said hardware (only) the option to increase the max_dB value in ALSA.

## Ensured volume levels for CKCv1 only at bootup

This change comes again in light of CKCv1 speaker. We limit the volume lowering to just the CKCv1 hardware at bootup.

**NOTE:** The base of this PR is only so such that we can review it easily.


radujipa:
[(3) As a user, I want to be able to make my speakers loud again when I get a replacement speaker [2]](https://trello.com/c/eKL9S6kO/1353-3-as-a-user-i-want-to-be-able-to-make-my-speakers-loud-again-when-i-get-a-replacement-speaker-2)

radujipa:
[(3) As a user, I want to be able to make my speakers loud again when I get a replacement speaker [2]](https://trello.com/c/eKL9S6kO/1353-3-as-a-user-i-want-to-be-able-to-make-my-speakers-loud-again-when-i-get-a-replacement-speaker-2)

debian/changelog
```diff
@@ -1,5 +1,12 @@
 kano-settings (4.0.2-0) unstable; urgency=low
```
skarbat: Increase version?

radujipa: You're right! I caught that one in this other PR: https://github.com/KanoComputing/kano-settings/pull/463/files

skarbat: Oh! Great!

PR #463, opened 2017-09-27
Changed copy in Settings > Display to include full screen model EDID

radujipa:
kano-screenshot-wed-sep-27-10-45-54

**NOTE:** The base of this PR is only so such that we can review it easily.


radujipa:
[(1) As a customer care agent, it would help to reveal EDID value in the about section [1]](https://trello.com/c/bKdUgzco/1342-1-as-a-customer-care-agent-it-would-help-to-reveal-edid-value-in-the-about-section-1)

PR #465, opened 2017-09-28
Removed legacy kano_screen_used config comment

radujipa:
After the resolution policy refactor, this comment is no longer
required at bootup. Essentially, it was used like an [EDID=] filter
at the time. Given that nothing uses it anymore, it's good to clean
up the config now.

PS: I can confirm that using `auto` in Settings > Display still works correctly.


radujipa:
[(8) [CARE] Screen Resolution issue when moving from TV to monitor. [8]](https://trello.com/c/vsFlFYaY/1172-8-care-screen-resolution-issue-when-moving-from-tv-to-monitor-8)

PR #466, opened 2017-10-02
Added EDID tools for dumping and flashing screen data

radujipa:
Two tools have been added here: rpi-dump-edid is a dev & user facing
script which dumps the screen EDID to the /boot/ partition for testing
data and debugging. generate-rpi-edid-dumps is a tool which uses raw
EDIDs to generate test data for kano-settings display functions.
These will be the tools to build on the res/edid dataset.

NOTE: Might be worth reviewing the `generate-rpi-edid-dumps` tool without the diff and as a whole file.


radujipa:
[Generate inferred EDID data from raw EDID dump for screen tests](https://trello.com/c/c6ss4uIO/1344-generate-inferred-edid-data-from-raw-edid-dump-for-screen-tests)

bin/rpi-dump-edid
```diff
@@ -0,0 +1,82 @@
...
+# The file is output to the /boot/ partition for easy access.
+#
+# This tool was used for all screen data under kano-settings/res/edid/raw
+# Requires sudo, RPi, and a screen.
+
+
+import os
+import sys
+
+from kano.utils.shell import run_cmd
```
skarbat: Just a question. What should be the general use here for `run_cmd`? `kano.utils.shell` or `kano.utils` ?

bin/rpi-dump-edid
```diff
@@ -0,0 +1,82 @@
...
+    if rc:
+        sys.stderr.write(
+            'Error: Could not calculate md5 sum for {}!\n rc = {}\n err = {}\n'
+            .format(tmp_edid_path, rc, err)
+        )
+        return RC_ERR_SUBPROCESS
+
+    md5sum = out.split(' ', 1)[0]
+    model_md5 = '{}-{}'.format(model, md5sum)
+
```
skarbat: Should this bit be wrapped in a `try-except` just in case?

tools/edid/generate-rpi-edid-dumps
```diff
@@ -5,51 +5,82 @@
...
+# Each raw EDID is flashed in turn to the attached display in order for the system
+# as a whole to think it's using the respective display. The original screen EDID will
+# be backed up and restored at the end (see ORIGINAL_EDID_BACKUP_PATH for the path).
+#
+# This tool requires the installation of edid-rw under /usr/bin.
+# https://github.com/bulletmark/edid-rw
+#
+# WARNING: This tool can destroy the machine and display!
+#          See edid-rw README for the dangers of using it.
+#
```
skarbat: Would it be better to move these important considerations into the docopt help section?

tools/edid/generate-rpi-edid-dumps
```diff
@@ -70,107 +101,363 @@ def run_cmd(cmd):
...
+
+    if not backup_original_edid():
+        return RC_ERR_SUBPROCESS
+
+    # Get all the raw EDID filenames without their extension.
+    # These will have been produced by the rpi-dump-edid tool.
+    screen_models_with_md5 = [
+        raw_edid[:-len(os.path.basename(raw_edid_path_pattern.format(filename='')))]
+        for raw_edid in os.listdir(raw_edid_dir)
+    ]
```
skarbat: nice

tools/edid/generate-rpi-edid-dumps
```diff
@@ -70,107 +101,363 @@ def run_cmd(cmd):
...
+            'Error: dtparam=i2c2_iknowwhatimdoing=on was not found in /boot/config.txt\n'
+        )
+        valid_config_txt = False
+
+    return valid_config_txt
+
+
+def check_i2c_bus(bus_number):
+    """ Check that the i2c device at address 3f805000.i2c is available. """
+
```
skarbat: This address is provided by `i2c_bcm2708` kernel module right?
Can we mention it here? Does it need a check using a `lsmod` or similar?

tools/edid/generate-rpi-edid-dumps
```diff
@@ -70,107 +101,363 @@ def run_cmd(cmd):
...
+        check_boot_config_txt() and
+        check_i2c_bus(RPI_HDMI_I2C_BUS_NUM)
+    )
+
 
 if __name__ == '__main__':
-    if os.geteuid() == 0:
-        print 'Program must NOT be run as root.'
-        sys.exit(RC_NOT_USER)
+    if os.geteuid() != 0:
```
skarbat: I am not sure that we need to `set-uid` this tool.

bin/rpi-dump-edid
radujipa: Looking back at it, the only reason why we have so many `from kano.utils import` statements is because we opted to write a nasty import there for backwards compatibility reasons. 

I think providing the full import path is clearer since you know at a glance where you need to look if you want to see the implementation of that function.

tools/edid/generate-rpi-edid-dumps
radujipa: An obscure one, I know... but reduces code maintenance

bin/rpi-dump-edid
radujipa: Good idea! Fixed

tools/edid/generate-rpi-edid-dumps
radujipa: Good point! Done

tools/edid/generate-rpi-edid-dumps
radujipa: Yes, did that one again... Good spot! Fixed

bin/rpi-dump-edid
skarbat: Nice, yeah I agree. 👍

tools/edid/generate-rpi-edid-dumps
radujipa: This one is required for the `edid-rw` tool to work properly on RPi. The i2c address to write to the HDMI port does not change. 

There are comments and a hint to use `i2cdetect -l` to check that it's available. What other checks would be necessary? (not sure I understand your questions)

skarbat: Oh I understand now. I wasn't sure about the purpose of the address that's all.

skarbat:
Great work, Radu.

radujipa:
Thanks for the review, @skarbat 👍

bin/rpi-dump-edid
```diff
@@ -0,0 +1,86 @@
...
+    out, err, rc = run_cmd('tvservice --name')
+    if rc:
+        sys.stderr.write(
+            'Error: Could not get the screen model!\n rc = {}\n err = {}\n'.format(rc, err)
+        )
+        return RC_ERR_SUBPROCESS
+
+    # If the EDID is corrupt, the proc might return '[E] No device present'.
+    if 'device_name' in out:
+        model = out.strip('device_name=').strip()
```
tombettany: Given that we almost certainly have a kano-settings function to do read the name, do we want to use that instead?

radujipa: Great point. That script started as a standalone tool, but forgot to integrate it with the system properly after I moved it to `bin/`. I'll look to move most of that functionality in kano-settings instead. Thanks for pointing that out 👍

radujipa:
@tombettany @skarbat 

As per Tom's suggestion, I removed the `rpi-dump-edid` and moved it into `kano-settings-cli` and the functionality in `kano_settings.system.display` instead.

PR #469, opened 2017-10-12
Adding a check in kano-wifi-gui for minimum length in passphrase

skarbat:
Provide a real-time check during passphrase input, which enables the Connect button only when the passphrase length is long enough.


skarbat:
[(8) As a user, I want helpful error messages when my wi-fi connection fails [2]](https://trello.com/c/v4m9mTUM/1332-8-as-a-user-i-want-helpful-error-messages-when-my-wi-fi-connection-fails-2)

kano_wifi_gui/PasswordScreen.py
```diff
@@ -141,7 +143,11 @@ def _on_connect(self, widget=None):
         )
 
     def _set_button_sensitive(self, widget, event):
-        self._connect_btn.set_sensitive(True)
+        '''
+        Enable the Connect button only if the passphrase is long enough
+        '''
+        passphrase_length=widget.get_text_length()
+        self._connect_btn.set_sensitive(passphrase_length >= self.passphrase_minlen)
```
radujipa: What happens when they are trying to connect to an Open network? 
Are we certain that all points have this min passphrase length of 8? WEP, WPA, WPA2?

radujipa: Also, should we show a message to the user about the min length such that it's clear why the button is disabled?

skarbat: You are right! WEP is either 5 or 13 in length. Discarding this PR and moving the logic to a lower layer.

PR #470, opened 2017-10-12
Fixed resetting Audio > Boost Volume setting on CKCv1

radujipa:
The issue here was that on CKCv1, kano-settings-onboot script set
the volume to low every time. The user is able to raise the volume
through Boost Volume option in the Audio menu which the bootup
script was not aware of, resetting the config. Created a system
file flag to note that the volume was lowered only once and the
user is responsible for the volume level thereafter.

**NOTE:** This fix is for the `3.13.0` `rc`.


radujipa:
[(3) As a user, I want to be able to make my speakers loud again when I get a replacement speaker [2]](https://trello.com/c/eKL9S6kO/1353-3-as-a-user-i-want-to-be-able-to-make-my-speakers-loud-again-when-i-get-a-replacement-speaker-2)

bin/kano-settings-onboot
```diff
@@ -139,21 +140,34 @@ def ensure_correct_browser(dry_run=False):
...
-    Sets the default max_dB option in ALSA for CKC only.
+    Sets the default max_dB option in ALSA for CKCv1 only once.
     """
+    was_set_once = os.path.exists(CKC_V1_ONBOOT_VOLUME_FLAG)
+    if was_set_once:
+        return
+
     is_ckc = is_ck2_pro(with_dbus=False)
+    if not is_ckc:
+        return
```
tombettany: Do you want to also create the flag here so that we don't have to run the slow `is_ck2_pro()` at every boot for all kits other than CKCv1?

radujipa: Given that the average runtime of that function is between `0.0008` - `0.0016` I don't think it's worth the added complexity. If we're going for performance, we should think about splitting this entire script into parallel units, I think.

PR #473, opened 2017-10-23
Reduce the time for checking Safe Mode hotkeys during bootup

radujipa:
Previously, the `kano-safeboot-mode` used `kano-keys-pressed` with
a retry 5 and sleep interval of 1s. This made the total execution
of the script around 8.7s. Reducing the sleep interval to just
10ms, the script should take around 5s. With the keypress interval
reduced, the resolution is improved making fast and slower presses
of the CTRL + ALT hotkeys easier to detect.


radujipa:
[(8) As a user, I would like my kit to boot up very quickly](https://trello.com/c/kgsbu9gZ/1301-8-as-a-user-i-would-like-my-kit-to-boot-up-very-quickly)

PR #475, opened 2017-10-30
Parental: Whitelist speaker claim page

tombettany:
The form used to claim the CKC speaker redirects to a page which would
be blocked by the ultimate parental settings. Prevent this from
happening by adding this URL to the whitelist.


tombettany:
[(1) As a CKC v1.0 user with ultimate parental settings enabled, I want to be able to access the claim your speaker page [1]](https://trello.com/c/DWsWggfc/1477-1-as-a-ckc-v10-user-with-ultimate-parental-settings-enabled-i-want-to-be-able-to-access-the-claim-your-speaker-page-1)

tombettany:
A note on this, the local copy of this file is only updated when the parental settings change so it will only actually unblock the site if the parental settings were enabled after we pushed this fix :-s

radujipa:
But then at least our response can be "please update"

radujipa:
Also, please remember to add a note in the `debian/changelog` for all these PRs.

tombettany:
Added changelog entry

PR #478, opened 2017-11-23
Provide a way to use a custom config.txt file for special screens

skarbat:
This changeset adds the capability to use a custom `config.txt` file in the boot partition, by placing a sentinel file. It is meant to be used for advanced users.

Normally, `kano-settings-onboot` would try to change to best settings and reboot the kit, but on unrecognized screens this can end up losing the intended settings and landing on a black screen, unable to recover.

What do you think @radujipa @Ealdwulf ?


radujipa:
[[P1] (5) Some TVs + monitors not working](https://trello.com/c/DSfAaHcz/1539-p1-5-some-tvs-monitors-not-working)

skarbat:
Thanks @radujipa , I have updated the README as well to remember how to handle special screens this way.

PR #479, opened 2017-12-05
Boot: Ensure LightDM configuration file exists

tombettany:
Ensure that the LightDM configuration exists on boot and regenerate if
needed.


bin/kano-settings-onboot
```diff
@@ -50,6 +50,9 @@ from kano_peripherals.wrappers.detection import is_ck2_pro, get_ck2_pro_version
 
 from kano_peripherals.wrappers.detection import CKC_V_1_1_0
 
+# FIXME: Kano Settings shouldn't really depend on Kano Init
+from kano_init.utils import ensure_lightdm_conf
```
Ealdwulf: Since this is a new function, technically you need to bump the required version of kano_init.

tombettany: That is exactly true, unfortunately we have a bunch of circular dependencies with this one and officially `kano-init` depends on `kano-settings`. In this case, it is a stand-alone script which is running this so we shouldn't cause any library problems but you bring up a point - I'll move the import into a `try` block so at least it doesn't kill everything. We need to revisit the whole `kano-settings` <-> `kano-init` setup to properly break all these dependency problems that we have.

tombettany:
[[P1] LXDE login screen error](https://trello.com/c/fBB8Bdgu/648-p1-lxde-login-screen-error)

PR #480, opened 2017-12-05
Write /boot/cpuid.txt

Ealdwulf:
This PR replaces https://github.com/KanoComputing/kano-toolset/pull/263
 Moved from kano-network-hook to avoid dependency on network.


bin/kano-settings-onboot
```diff
@@ -125,6 +129,25 @@ def main(args):
 
     return rc or RC_SUCCESS
 
+def write_cpu_id():
+    # Get the unit CPU Serial number and send it, save a local copy
+    cpuid = get_cpu_id()
+    cpuid_filename_boot = '/boot/cpuid.txt'.format(cpuid)
```
radujipa: Missing `{}` somewhere here?

Ealdwulf: No, we don't want the cpuid in the filename, unlike the one in `/etc`. That's a no-op, I'll remove it.

skarbat: I think we want to remove the format(), to keep a fixed filename?

Ealdwulf: Fixed.

PR #482, opened 2017-12-14
Add s3 repo mirror.

Ealdwulf:
Add s3 repo mirror to parental control whitelist.


radujipa:
Also, shouldn't this be against rc?

Ealdwulf:
@radujipa It should have been against rc, however it turns out that all of amazonaws.com is already in the whitelist. So this is mainly for clarity, and I will leave it until after the release. 
Will add a changelog entry before merging (when we decide what the next version number is).

PR #486, opened 2017-12-18
Fixed long delay when going into the Notifications menu

radujipa:
The issue here was that every time when you accesssed the
Notifications menu there was a long delay (up to 10s or more)
due to the UI retrying for D-Bus ifaces for PiHat and LED Speaker.


radujipa:
[As a user, I want to be able to access the Settings > Notifications menu quickly](https://trello.com/c/bmMhpvZ9/1696-as-a-user-i-want-to-be-able-to-access-the-settings-notifications-menu-quickly)

PR #487, opened 2017-12-18
Propagate changes to WHITELIST from master.

Ealdwulf:
Propagate #484 + #482 to jessie


Ealdwulf:
@tombettany I think we shoudl have merged this. Maybe it's causing some update failures?

tombettany:
Yeah, I think that you are right, merge away.

PR #490, opened 2018-01-04
Make safe mode use the preferred mode of the screen. NOT FOR MERGE

Ealdwulf:
This is the proposed fix for the hotfix image. 

NB It is NOT proposed that we merge this until we have thought more.


radujipa:
[As a user, I want Safe Mode to use the screen preferred resolution](https://trello.com/c/1291Cx7d/1716-as-a-user-i-want-safe-mode-to-use-the-screen-preferred-resolution)

PR #491, opened 2018-01-08
Safe Mode fixes for 3.14.1 hotfix image

radujipa:
I've cherry-picked the changes discussed for the 3.14.1 image and narrowed the down to these commits only. 
The PR is against `rc` since it's the same as the release branch currently.

It shouldn't be too much hassle to rebase master on top of `rc` afterwards.


tombettany:
@skarbat That is because unfortunately this repo got bumped to 4.x ages ago and so we just have to deal with it until our version catches up with this guy.

radujipa:
[As a user, I want Safe Mode to use the screen preferred resolution](https://trello.com/c/1291Cx7d/1716-as-a-user-i-want-safe-mode-to-use-the-screen-preferred-resolution)

radujipa:
[As a user, I do not want to have hdmi_drive=2 enabled in Safe Mode](https://trello.com/c/6oYN2Ldy/1717-as-a-user-i-do-not-want-to-have-hdmidrive2-enabled-in-safe-mode)

PR #494, opened 2018-01-24
Fixed applying config when clicking on Flip Screen option

radujipa:
The issue was that the configuration, display_rotate=0/2, was not
being applied at the end, when clicking on Apply Changes button.
This would confuse users to the actual state of the setting and
slowed down the UI when clicking on the CheckButton.


radujipa:
[[P2] Settings > Display > Flip Screen sets config on click rather than on apply](https://trello.com/c/U1GVDtmN/1734-p2-settings-display-flip-screen-sets-config-on-click-rather-than-on-apply)

PR #495, opened 2018-01-29
Adapted the call to kano-feedback-cli after usage refactor

radujipa:
This simply updates the call to the binary with the new usage
interface.

Related to: https://github.com/KanoComputing/kano-feedback/pull/85


radujipa:
[(1) As an engineer, I want to change the error messages for V30DB so I can learn more about what's going on [2]](https://trello.com/c/fVSfXXqG/1747-1-as-an-engineer-i-want-to-change-the-error-messages-for-v30db-so-i-can-learn-more-about-whats-going-on-2)

bin/kano-capture-logs
```diff
@@ -25,8 +28,13 @@ def store_logging_error():
 
 
 def get_logs():
-    rc = os.system("USER=root LOGNAME=root HOME=/root kano-feedback-cli {} >{} 2>&1".format(
-        LOG_DEST, LOG_OUTPUT))
+    rc = os.system(
+        'USER=root LOGNAME=root HOME=/root'
+        ' kano-feedback-cli --output {} --screenshot'
```
Ealdwulf: Given this happens at boot, I'm not sure it's worth doing a screenshot - on the other hand it's also pretty harmless.

radujipa: It used to be there, but the `--title` tag will be more useful to know where it's coming from. Leaving it there obviously makes it slower to bootup into Safe Mode, and the archive a bit bigger. I'll remove it 👍

bin/kano-capture-logs
```diff
@@ -38,13 +45,16 @@ def get_log_size():
...
+
+    get_logs()
 
-if (get_boot_space()-get_log_size()) > SPACE_MARGIN:
-    os.system("cp {} /boot".format(LOG_DEST))
-    os.system("syncfs </boot/config.txt")
-    play_sound(SOUND_FILE)
+    if (get_boot_space() - get_log_size()) > SPACE_MARGIN:
+        os.system("cp {} /boot".format(LOG_DEST))
+        os.system("syncfs </boot/config.txt")
```
skarbat: Is this `syncfs` meant to flush `config.txt` or the logfile?

Ealdwulf: I don't know which is intended, but it applies to the whole filesystem so it will do both.

PR #497, opened 2018-02-16
Self problems: Wi-Fi doesn't timeout when I/F is down

wizofe:
Fixing a bug caused when a Wi-Fi I/F is removed (for example a hotspot when trying to log in via WPA) caused due to an undefined global reference exception.


kano_settings/system/display.py
```diff
@@ -306,6 +306,9 @@ def set_safeboot_mode():
     set_screen_value('hdmi_drive', None)
     set_config_value('hdmi_drive', None)
 
+    set_screen_value('hdmi_drive', None)
+    set_config_value('hdmi_drive', None)
+
```
skarbat: Perhaps you sent `display.py` by mistake?

PR #499, opened 2018-02-23
Remove call to dashboard-onboot after its removal

radujipa:
Requires: https://github.com/KanoComputing/os-dashboard/pull/372


radujipa:
[(5) As a user who unplugs during an upgrade, I want the updater to relaunch itself to finish](https://trello.com/c/IE04dtnM/1766-5-as-a-user-who-unplugs-during-an-upgrade-i-want-the-updater-to-relaunch-itself-to-finish)

PR #500, opened 2018-02-27
Add a flag in the debian/rules to stop rewriting shebang lines

radujipa:
It seems like without this magic flag combined with `--with python2` `debuild` thinks it's a grand idea to remove my shebang by default. 👎 

It turns out to be the reason why [this splash PR](https://github.com/KanoComputing/kano-settings/pull/489) failed after a build...

I'm going to find me one of these deb devs one day 😈


radujipa:
[(1) As a user, I want a loading animation when I launch Settings app](https://trello.com/c/6COcubhe/1702-1-as-a-user-i-want-a-loading-animation-when-i-launch-settings-app)

wizofe:
By the way I know on of those. I can introduce you to him 🤗

PR #501, opened 2018-03-14
Fix deb version of PAM all caps naming

radujipa:
When this module was transitioned from pip (0.1.4) to deb (0.4.2-13.1),
the version in Raspbian contained an uppercase name for the module.
This method of importing fixes this and accounts for the pkg being
fixed in the future to be lowercase.


wizofe:
That is if we are sure we are transitioning from asking for pinned versions to any `>=` versions though. As the pinned version is still: 

```
kanux@cc548b1c3291:~$ dr list versions python-pam
dr WARN This subcommand is deprecated, please use builds instead
dr info Listing all builds of python-pam
dr info   0.1.4
```
Which could potentially cause problems, thinking, if the packages aren't 100% compat?

PR #505, opened 2018-04-17
Fixed issue with finding optimal resolutions for certain screens

radujipa:
The issue here was when trying to access the list of optimal
resolutions for screens with aspect ratios different from 4:3,
16:9, and 16:10.


radujipa:
[[P3] Screens with 5:4 aspect ratios cannot set an optimal resolution](https://trello.com/c/wV3Y5tQe/1849-p3-screens-with-54-aspect-ratios-cannot-set-an-optimal-resolution)

PR #511, opened 2018-06-21
Use old audio driver to avoid kernel / firmware crash

radujipa:
According to https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=195178
it would seem that there was a newer version of the audio driver
introduced and used in Raspbian since 11 Oct 2017. This however
clashes under heavy GPU load when sound is playing as documented here
https://github.com/raspberrypi/linux/issues/2587 As such, revert to
use the old one for added safety.

---

If I remember correctly, we haven't updated our sources since 11 Oct 2017 (need to verify this on 3.15.0). This means that the new audio driver is something else that was introduced when we started upgrading to stretch. We would need to test the quality and regressions if we decide to merge this. Using this option with the old qmlmatrix prevented the crashes we've previously seen.

### DIFF
--- a/boot_default/config.txt
+++ b/boot_default/config.txt
@@ -76,6 +76,11 @@ display_rotate=0
 # Decrease audio noise.
 disable_audio_dither=1
 
+# Disable the new audio pwm driver to avoid the risk of a kernel / firmware crash.
+# https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=195178
+# https://github.com/raspberrypi/linux/issues/2587
+audio_pwm_mode=1
+
 # ---------------------------------------------------------------------------------------
 # Configuration Filters - RPis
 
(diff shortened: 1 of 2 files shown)


skarbat:
Looks very reasonable, nice catch! I have a question though... is `=1` the appropriate one for us?
There seem to be 2 other values, `=2` being a better one apparently:

```
audio_pwm_mode=1 -> for when only some simple sound is required (beeps & such)
audio_pwm_mode=2 -> good quality & better compatibility with the BCM2835 SoCs (actual issue)
audio_pwm_mode=3 -> improved quality (subjective for my ears, I don't own an oscilloscope/spectrum analyzer) - the new algorithm for Pi2B&Pi3B&above. Although, I haven't yet done any DVB-S2 MPEG4 1080p - 50FPS tests on the Pi2B boards I have with this new algorithm
```

 * https://forum.kodi.tv/showthread.php?tid=326206

radujipa:
Good find @skarbat ! I had some issues finding some documentation about this option, so I suspected the default before upgrading to stretch would've been =1. I agree that we'd need to test the audio quality with this option as well.

PR #512, opened 2018-06-22
Fixed Safe Mode configuration sometimes not getting applied correctly

radujipa:
The issue seems to have been caused by the SysV services kano-safeboot
and kano-settings not being ordered correctly.

NB: The Trello ticket suggested a much needed and larger set of refactors with this, but given the `rc` phase, I've made the minimal amount of changes for this.


radujipa:
[[P2] Safe Mode configuration not applied consistently after reboot](https://trello.com/c/L6s2fDi2/1822-p2-safe-mode-configuration-not-applied-consistently-after-reboot)

PR #513, opened 2018-07-04
Reboot on account creation

Ealdwulf:
This PR changes kano-settings so that  adding an account, requires a reboot immediately, but offers the user the opportunity to back out and close their files.


kano_settings/set_account.py
```diff
@@ -93,11 +93,26 @@ def go_to_password_screen(self, widget, event):
             self.win.clear_win()
             SetPassword(self.win)
 
+    def add_account_dialog(self):
+            kdialog = kano_dialog.KanoDialog(
```
radujipa: Double indentation in this function?

Ealdwulf: Fixed

kano_settings/set_account.py
```diff
@@ -93,11 +93,26 @@ def go_to_password_screen(self, widget, event):
             self.win.clear_win()
             SetPassword(self.win)
 
+    def add_account_dialog(self):
+        kdialog = kano_dialog.KanoDialog(
+            _("Add an Account"),
+            _("To add an account, you need to reboot. If you have any files open, you should cancel and save your files first. Add and reboot now?"),
+            [
+                {"label": _("ADD and REBOOT NOW"), "return_value": True, "color": "orange"},
```
radujipa: From @brandonjackson: Please change button label to "REBOOT NOW"

PR #514, opened 2018-07-05
Fix traceback when bluetooth didn't start

Ealdwulf:
This PR make sure access to the bluetooth dbus is not required to start kano-settings.
Some objects were created at module init time, so prevented it starting properly.
Also fix unicode error.


Ealdwulf:
This is now ready. 
It turns out the UI was crashing an nearby  device's name had a unicode character (such as, for example, 'AJB’s MacBook Pro')

PR #517, opened 2018-08-13
[OS-232] Removed default RPi2 CPU frequency settings from config.txt

radujipa:
By removing these configuration options, RPi3 will no longer need
to reboot on a first ever boot. Additionally, it means that the
options themselves will now be handled by the RPi firmware.


wizofe:
Looks good.

Talking about that, do we want to (test with and) assign more memory as well? Like `320`, as it's used on MM Distros like KODI?

radujipa:
@wizofe The memory limits are a bit of a delicate change that we'd only do only if we'd need to...

PR #519, opened 2018-09-17
Screen Kit mechanism to ensure display orientation consistency

radujipa:
Reviewable preliminary version of the PR, currently testing the approach.

## Ensure display_rotate is set globally for Screen Kits
When users (un)tick the Display > Flip option when using SKs, the
display_rotate config option is mirrored globally. This is to prevent
situations when SKs lose EDID that the screen orientation changes.

## Ensure screen kit orientation is set globally on first boot
The display_rotate option is already set for all SKs in the config
file using EDID filters. This ensures that the global setting also
matches the screen rotation for SKs to ensure consistency in the
unlikely event an SK loses EDID.


bin/kano-settings-onboot
```diff
@@ -217,6 +219,22 @@ def ensure_correct_volume_levels(dry_run=False):
...
 
+def ensure_screen_kit_rotation_global():
+    """
+    Ensure the screen kit orientation setting is also set globally.
+
+    This sets a 'fallback flag' in the unlikely situation where a SK will
+    lose its EDID data. No reboot is required since the option is already set
+    in the config.txt by default - this just mirrors it.
+    """
+    if not get_config_value('display_rotate') and is_screen_kit():
```
Ealdwulf: `is_screen_kit` uses EDID. What if that's already corrupted?

radujipa: If the EDID is not there even at first boot, I'm throwing in the towel. It shouldn't pass manufacturing QA!

Ealdwulf: As per verbal discussion - there is - we could check the USB.

bin/kano-settings-onboot
```diff
@@ -217,6 +222,26 @@ def ensure_correct_volume_levels(dry_run=False):
...
+    if was_set_once:
+        return
+
+    if is_screen_kit():
+        rotation = get_screen_value('display_rotate')
+        set_config_value('display_rotate', rotation)
+
+        if not os.path.exists(SYSTEM_FLAGS_DIR):
+            os.makedirs(os.path.dirname(SYSTEM_FLAGS_DIR))
+        open(DISPLAY_ROTATE_GLOBAL_FLAG, 'a').close()
```
tombettany: Maybe wrap in a try in case permissions or something causes this to traceback and not run the rest of the script.

radujipa: Wrapped the whole function 👍

PR #520, opened 2018-09-27
Fix inconsistent Resolution: Auto setting for unflashed screens

radujipa:
Currently, the config.txt global hdmi_group/mode contain settings
specific for Kano screens in case their EDID gets corrupted. However,
these settings get wiped out when selecting Auto in Settings > Display.
Handle Auto for corrupted EDIDs separately and keep the previous
behaviour for the rest. The decision was made to prioritise Kano
SKs in the corrupted EDID case, with the caveat that other screens
might not work. The change was not reflected in kano-settings-onboot
such that users might still workaround this manually should they
need to.


kano_settings/system/display.py
```diff
@@ -373,6 +373,28 @@ def is_screen_kit(use_cached=False):
...
+    These are specific to the Kano screens.
+
+    Returns:
+        bool - Whether or not changes were made and a reboot is required.
+    """
+    hdmi_group = 2
+    hdmi_mode = 28
+
+    changes_required = (
+        get_config_value('hdmi_group') != hdmi_group and
```
tombettany: Surely this should be `or`. The value needs to be updated if either of the values are incorrect.

radujipa: Rats, good catch! Thanks!

PR #521, opened 2018-09-27
 [OS-366, OS-368] config.txt: Fix corruption detection

tombettany:
The `config.txt` corruption detection method was checking for the
existence of a parameter which was removed from the default
configuration which meant that it would always claim that the file had
become corrupted. Resolve this by using a parameter (`dtparam`) which
will always be present.

Also fixes logging tracebacks in kano-settings-onboot:

If the boot configuration was being detected as corrupted, a logging
message would be written but this would fail as the `warning()` method
wasn't defined on the `logger` object. Resolve this by switching to use
the defined `warn()` message.


boot_default/config.txt
```diff
@@ -57,6 +57,11 @@ start_x=0
 # Note: RaspberryPI model B+ only
 max_usb_current=0
 
+
+# Note: kano_settings.boot_config.check_corrupt() checks for presence of
+#       "dtparam" to determine if this file is corrupt, don't remove without
+#       updating that function
```
radujipa: At a minimum I was thinking we'd have tests to verify that what the function was looking is in this file. 
But I guess this is the next minimum...

kano_settings/boot_config.py
```diff
@@ -104,14 +104,22 @@ def check_corrupt(self):
...
         found = set()
 
         for l in lines:
             for m in must_contain:
                 if m in l:
                     found.add(m)
-        return must_contain != found
+
+        if must_contain == found:
+            return False
```
radujipa: The implementation here is fundamentally flawed; users should be free to change anything in the config.txt file without our libs breaking / resetting. I suggest the function is removed completely.

tombettany: Right now this is the only way that we know if the file has become corrupted. I'm happy for you to create ticket to remove this but at this point I'm not changing the fundamental implementation to allow for users to arbitrarily modify the `config.txt` (which we randomly replace anyway)

skarbat: Just for to keep in mind for later, I already implemented a mechanism to achieve this limitation, explained here:

 * https://github.com/KanoComputing/kano-settings/#display

and implemented here:

 * https://github.com/KanoComputing/kano-settings/blob/master/etc/init.d/kano-settings#L27

PR #522, opened 2018-10-08
Fixed an issue in Parental Controls when changing config levels

radujipa:
When switching from Ultimate to High and then selecting Configure
the changes would not be applied when switching between views. This
meant that the Apply Changes in the Configure allowed/blocked only
set the changes for that view. Hence the level would remain on
Ultimate rather than High.


radujipa:
Comments still welcome!

PR #524, opened 2018-10-09
Closing config transaction during onboot script to avoid tbk

radujipa:


radujipa:
Comments still welcome!

PR #528, opened 2018-12-28
[OS-728] - Temporary fix to restoring wireless connection on boot

skarbat:


skarbat:
Definitely agreed @tombettany - Providing a cleaner solution using the `nohook` keyword.
The functionality fixes the problem at my end after several reboot and reconnection tests.

Need to test that the packaging, `dpkg-divert` works flawlessly on image build and upgrades.
What are your thoughts on that strategy?

skarbat:
This PR is obsoleted by https://github.com/KanoComputing/kano-settings/pull/529
Closing now.

PR #536, opened 2019-09-16
Bump changelog and package version

radujipa:
Image build is failing because the changelog and version were not bumped: Overture and Updater are depending on 4.3.3.


tombettany:
👍

